PR #25, opened 2020-03-02
Add ReadWriteRef

sirmax:


core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -0,0 +1,160 @@
...
+  def write: Resource[F, ReadWriteRef.Upd[F, A]]
+}
+
+object ReadWriteRef {
+  /**
+   * A "resource" of [[ReadWriteRef.write `write`]].
+   * `(A => F[A])` part lets you read current value and produce a new one, effectfully.
+   * `F[A]` in the return part lets you use new `A` in the outer context, e.g. return it from `use`.
+   */
+  trait Upd[F[_], A] extends ((A => F[A]) => F[A])
```
t3hnar: why `trait` and not `type` ?

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -0,0 +1,160 @@
...
+
+  object Upd {
+    implicit final class SelfOps[F[_], A](val self: Upd[F, A]) extends AnyVal {
+      /** A shorthand for non-effectful updates. */
+      def plain(f: A => A)(implicit F: Applicative[F]): F[A] = self(a => F.pure(f(a)))
```
t3hnar: `def map`?

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -0,0 +1,160 @@
...
+
+      /** A shorthand for non-effectful update that ignores previous value. */
+      def const(a: A)(implicit F: Applicative[F]): F[A] = self(_ => F.pure(a))
```
t3hnar: `def set` ?

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -0,0 +1,160 @@
...
+      // and notifies pending listeners.
+      //
+      // IMPORTANT:
+      //   This function must be called only from uncancellable blocks (e.g. Resource release).
+      def release(a: A, useChange: Int => Int): F[Unit] = stateRef
+        .modify { s0 =>
+          val inUse = useChange(s0.inUse)
+
+          s0.pendingWrites match {
+            // Give writers a priority
```
t3hnar: why do we change the order?

I think calls should be executed in the same order as originally made and not just we do writes first and there after proceed with reads

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: For extension methods. They can't be resolved for `type`.

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: Hmm… no. `map` would be `def map[B](f: A => B): Upd[F, B]`.

I called it `pure` at first but got clashes with cats `Applicative` and changed it to `plain`. Any better names for something "non-side-effectful"?

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: Maybe, but again, it's a method on an "updater" so it kinda reads "update with a constant". Still, `set` might fit just fine.

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: This is not about changing the order. It's more about throughput. Readers execute all-at-once, writers – one-by-one. They have separate "queues" for that exact reason.

My idea of pushing writers forward was
- to let readers have more up-to-date value
- letting readers pile-up and then getting rid of them in one go

I can imagine that in pathological cases writers can overwhelm readers, but perhaps we can optimise for that case a bit later.

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: Renamed in 6be0e96.

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
t3hnar: `bind`?

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
t3hnar: I think this might be very confusing in some cases, when execution happens in other than call order.

sirmax: But `ReadWriteRef` is all about re-ordering in a sense that writers have to wait while reads are running, and more reads may get ahead of them. That's the difference between this one and `SerialRef`.

t3hnar: the case I'm talking about is for instance you made a `write` and there after `read`, and you may end up in situation when `read` overtook `write`

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: No, I don't like `bind` either. Neither the function in question matches Haskell `bind` which is synonymous to `flatMap`, not it has plain English "bind" semantics.

t3hnar: ok, I'm let's keep the current one

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
sirmax: "After" implies a monadic sequence in which case `read` won't even be attempted before `write`. And without such sequencing you have a concurrent execution where anything can happen ;)

sirmax: Buuut, speaking of concurrent execution and ordering, let's focus on another piect of code which we may want to improve.

Right now we execute pending reads like this
```scala
val effect = s0.pendingReads.traverse_(l => l(a))
```
And `pendingReads` is a `List` which is built by _prepending_ late readers…
So we have reordering and _sequential_ execution. Should we improve both? If yes – how?

sirmax: The most obvious would be to reverse this list (or use a `Queue`) and do `parTraverse_`. Do you see any drawbacks here?

t3hnar: > "After" implies a monadic sequence in which case read won't even be attempted before write. And without such sequencing you have a concurrent execution where anything can happen ;)

```scala

val promise: ...
read.use { promise.get }.toFuture

write.use.toFuture
read.use.toFuture

promise.complete()
```

I know, that's an extreme case :)

sirmax: Sure, but that exactly shows how `ReadWriteRef` works – the second `read.use(…).toFuture` will execute in parallel with the first read and has all the chances to finish first.

t3hnar: I would do stuff like change a state a bit - so you have `state:List[Action]` where
```
sealed trait Action
object Action {
 case class Reads(value: List[A => F[Unit]])
 case class Write(value: A => F[Unit])
```

So this way you collect reads in between writes and execute them in `par` (you can do that with Parallel or still remain with Concurrent)

t3hnar: writes are executed strictly serially

t3hnar: regardless of this discussion I can accept this PR. Just wanted to make a point here

sirmax: Hmm this would change the execution model considerably. Right now readers can execute immediately as long as there is no write running.

If I get you right, you suggest blocking readers as soon as there is a writer _pending_. It's not wrong, I can even agree that it's more fair in its strictness. Okay, let me think of it a bit.

t3hnar: I even think on renaming a bit for clarity 

```
sealed trait Serial
object Serial {
 case class Parallel(reads: List[A => F[Unit]]) extends Serial
 case class Write(value: A => F[Unit]) extends Serial
```

sirmax: Reimplemented in acf2ba6e.

core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -0,0 +1,193 @@
...
+            case PendingWrite(listener) +: rest if inUse == 0 =>
+              val s1 = s0.copy(a = a, inUse = -1, pending = rest)
+              val effect = listener(a).start.void
+              s1 -> effect
+
+            case _ =>
+              // A slightly optimized way to split the pending ops in two and start the first batch
+              @tailrec def startPending(effect: F[Unit], n: Int, ps: Queue[Pending]): (F[Unit], Int, Queue[Pending]) = {
+                ps match {
+                  case PendingRead(listener) +: tail => startPending(effect <* listener(a).start, n + 1, tail)
```
t3hnar: does that `match` work with `dequeue` underneath ?

sirmax: It should, but let me check it.

sirmax: Hah! It does `(head, tail)` and while `tail` is quite efficient, `head` may be sub-optimal in some cases. Let me improve that.

sirmax: Done in e87a7bc.

PR #29, opened 2020-04-05
NonEmptyList partitioning

mr-chillman:
When splitting `cats.data.NonEmptyList[T]` to two collections we can turn it `toList` and then `split` but that would not give us compilation level guarantee that one of those resulted collections will be non-empty.
To fix that `NonEmptyListPartitions[L,R]` is introduced. `NonEmptyListPartitions[L,R]` can be either `AllLeft[L,R]` either 'AllRight[L,R]` or `Both[L,R]`

Usage:
```
import NonEmptyListPartitionOps._

NonEmptyList.of(1, 2, "s", "r").split({
   case i: Int    => Left(i)
   case s: String => Right(s)
}) shouldBe Both(
   NonEmptyList.of(1, 2),
   NonEmptyList.of("s", "r")
)
```

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/NonEmptyListPartitionOps.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/NonEmptyListPartitionOps.scala
@@ -1,0 +1,51 @@
+package com.evolutiongaming.catshelper
+
+import cats.data.NonEmptyList
+import com.evolutiongaming.catshelper.NonEmptyListPartitions.{AllLeft, AllRight}
+
+object NonEmptyListPartitionOps {
+
+  implicit class NonEmptyListPartitionOpsC[T](nel: NonEmptyList[T]) {
+    def split[L, R](classifier: T => Either[L, R]): NonEmptyListPartitions[L, R] = {
+      val value: NonEmptyListPartitions[ L, R] = classifier(nel.head).fold(AllLeft.one, AllRight.one)
+
+      nel.tail.foldLeft(value) { case (acc, nextElement) =>
+        classifier(nextElement).fold(
+          acc.appendLeft,
+          acc.appendRight
+        )
+      }
+    }
+  }
+}
+
+sealed trait NonEmptyListPartitions[L, R] {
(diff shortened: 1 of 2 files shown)


sirmax:
There is already [`Reducible.nonEmptyPartition`](https://typelevel.org/cats/api/cats/Reducible.html#nonEmptyPartition[A,B,C](fa:F[A])(f:A=%3EEither[B,C]):cats.data.Ior[cats.data.NonEmptyList[B],cats.data.NonEmptyList[C]]) ;)

mr-chillman:
> There is already [`Reducible.nonEmptyPartition`](https://typelevel.org/cats/api/cats/Reducible.html#nonEmptyPartition%5BA,B,C%5D(fa:F%5BA%5D)(f:A=%3EEither%5BB,C%5D):cats.data.Ior%5Bcats.data.NonEmptyList%5BB%5D,cats.data.NonEmptyList%5BC%5D%5D) ;)


Oh well. That was exactly what I was looking for but failed to find :c
Thanks!

sirmax:
@mr-chillman to be honest, I have just found it myself. Here's how:
- first thought was, do we need a custom ADT when there's `cats.data.Ior`?
- next, can we make it work for other non-empty data types like `NonEmptyChain`?
- checking sources for "non empty" typeclasses in `NonEmptyList`…
- found `NonEmptyTraverse`, cool, can we use its `nonEmptyTraverse` to implement this?
- …hmm… nope, but what else is there?
- stumbled upon `nonEmptyPartition` – hey, that's exactly what we're implementing! :D

mr-chillman:
> @mr-chillman to be honest, I have just found it myself. Here's how:
> 
> * first thought was, do we need a custom ADT when there's `cats.data.Ior`?
> * next, can we make it work for other non-empty data types like `NonEmptyChain`?
> * checking sources for "non empty" typeclasses on `NonEmptyList`…
> * found `NonEmptyTraverse`, cool, can we use its `nonEmptyTraverse` to implement this?
> * …hmm… nope, but what else is there?
> * stumbled upon `nonEmptyPartition` – hey, that's exactly what we're implementing! :D

Didn't know about `Ior` hehe

PR #40, opened 2020-08-05
Fix cancellation issues in ReadWriteRef

### LINKED ISSUE #39: Cancelling ReadWriteRef.read can result in a locked-up state
Initially we observed this issue with `FeatureToggled`. It went like so:
- we have a feature-toggled resource, toggled on and under some use
- we toggle it off
- later we toggle it on, but it does not go back.
Heap dumps showed that an internal `rwLock: ReadWriteRef` had a positive number of active tasks despite them being complete long time ago.

We've narrowed down the problem to the incorrect cancellation handling in `ReadWriteRef`, specifically this bit
```scala
val read: Resource[F, A] = {
  Resource.suspend {
    Concurrent.cancelableF[F, Resource[F, A]] { cb =>
      …
      stateRef
        .modify { // acquisition
            …
            val effect = … // emit a Resource that modifies stateRef on release
            val cancel = F.unit
            s1 -> (effect as cancel)
        }
        .flatten
    }
  }
```
In a nutshell, we have two `stateRef` updates here: one for "acquisition" and one for "release", but the problem is that they happen on different "layers", and its quite possible that only "acquisition" happens but not "release" when the `read.use(…)` gets cancelled.

sirmax:
Fixes #39

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
@@ -5,6 +5,7 @@ import cats.effect.{Concurrent, Resource}
 import cats.effect.concurrent.{Deferred, Ref}
 import cats.effect.implicits._
 import cats.implicits._
+import com.evolutiongaming.catshelper.CatsHelper._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Queue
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/evolutiongaming/catshelper/ReadWriteRef.scala
```diff
@@ -71,124 +71,166 @@ object ReadWriteRef {
...
-      val write: Resource[F, Upd[F, A]] = {
-        Resource.suspend {
-          Concurrent.cancelableF[F, Resource[F, Upd[F, A]]] { cb =>
-            val listener = (a: A) => {
-              val r = Resource {
-                Ref[F].of(a) map { aRef =>
-                  val upd: Upd[F, A] = (f: A => F[A]) => aRef.get.flatMap(f).flatTap(aRef.set)
-                  val commit = aRef.get.flatMap(a => release(a, _ => 0))
-                  upd -> commit
+    (Ref[F].of(init), Ref[F].of(State())) mapN { (aRef, stateRef) =>
```
t3hnar, line 138: could you please explain on why you need two separate Refs ?

sirmax: Separation of concerns mostly. I tried to keep the `stateRef` only for read/write coordination to make it easier to understand (at least for myself).

Probably we can fit `a: A` back into the `State` but I'm not sure I want to do this right now. Do you see any problems with a separate ref?

PR #45, opened 2020-09-25
Bump cats version.

saraiva132:
Bumps:

- cats
- cats-effect
- scalatest

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/DataHelper.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/DataHelper.scala
@@ -9,7 +9,10 @@ object DataHelper {
 
   implicit class SortedMapOpsDataHelper[K, V](val self: SortedMap[K, V]) extends AnyVal {
 
-    def toNem: Option[Nem[K, V]] = Nem.fromMap(self)
+    @deprecated("no longer required by cats" , since = "2.1.1")
+    def toNem(implicit order: Order[K]): Option[Nem[K, V]] = Nem.fromMap(self)
+
+    def toNem(): Option[Nem[K, V]] = Nem.fromMap(self)
   }
 
 
@@ -25,7 +28,7 @@ object DataHelper {
     def toNem(implicit order: Order[K]): Option[Nem[K, V]] = {
       self
         .toSortedMap
-        .toNem
+        .toNem()
     }
   }
 
@@ -48,7 +51,7 @@ object DataHelper {
       self
         .toSortedMap
         .map { case (k, v) => f(k, v) }
-        .toNem
+        .toNem()
         .get
     }
 
@@ -57,7 +60,7 @@ object DataHelper {
       self
         .toSortedMap
         .map { case (k, v) => (f(k), v) }
-        .toNem
+        .toNem()
         .get
     }
   }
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/evolutiongaming/catshelper/DataHelper.scala
```diff
@@ -9,7 +9,7 @@ object DataHelper {
 
   implicit class SortedMapOpsDataHelper[K, V](val self: SortedMap[K, V]) extends AnyVal {
 
-    def toNem(implicit order: Order[K]): Option[Nem[K, V]] = Nem.fromMap(self)
+    def toNem: Option[Nem[K, V]] = Nem.fromMap(self)
```
t3hnar, line 12: it's not backward compatible change, you can add one more method and deprecate old one

core/src/test/scala/com/evolutiongaming/catshelper/ResourceBreakFlatMapChainTest.scala
```diff
@@ -18,17 +18,17 @@ class ResourceBreakFlatMapChainTest extends AsyncFunSuite with Matchers {
     val result = ()
       .pure[IO]
       .toResource
-      .multiply(100)
```
t3hnar, line 21: please put 100 back to not have long running time

core/src/test/scala/com/evolutiongaming/catshelper/ResourceBreakFlatMapChainTest.scala
```diff
@@ -18,17 +18,17 @@ class ResourceBreakFlatMapChainTest extends AsyncFunSuite with Matchers {
...
       .attempt
       .void
 
     FromFuture[IO]
       .apply { result.toFuture }
-      .timeout(1.second)
+      .timeout(3.second)
       .attempt
-      .flatMap { a => IO { a should matchPattern { case Left(_: TimeoutException) => } } }
+      .flatMap { a => IO { a should matchPattern { case Right(()) => } } }
```
t3hnar, line 30: @sirmax looks like https://github.com/typelevel/cats-effect/issues/469 is fixed now

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/DataHelper.scala
saraiva132: Yeah. thanks

project/plugins.sbt
```diff
@@ -8,4 +8,6 @@ addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.2.7")
 
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.13")
 
-addSbtPlugin("com.evolutiongaming" % "sbt-scalac-opts-plugin" % "0.0.5")
\ No newline at end of file
+addSbtPlugin("com.evolutiongaming" % "sbt-scalac-opts-plugin" % "0.0.5")
+
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.5.1")
```
t3hnar, line 13: why do we need this here? @saraiva132 you could just add that in your local `.sbt/` folder

saraiva132:
I think it's a super useful plugin. I can remove if you want, for this
project, might not make sense. Remove?

project/plugins.sbt
saraiva132: I removed it ;)

t3hnar: it is indeed super useful! hence just add it locally to have for all projects, the idea is that you don't add optional stuff into project files, but rather into your local config :)

PR #75, opened 2021-04-01
Failed attempt to migrate to cats effect 3

rtar:
I could not migrate tests because of custom thread executors, but adding a draft to, may be, work on it later, an shot it off.


core/src/main/scala/com/evolutiongaming/catshelper/EffectHelper.scala
```diff
@@ -1,18 +1,18 @@
...
 import scala.concurrent.Future
 import scala.util.Try
+import cats.effect.{ Deferred, MonadCancel }
 
 @deprecated("use CatsHelper instead", "1.0.1")
 object EffectHelper {
 
   @deprecated("use BracketOpsEffectHelper", "1.0.0")
-  class EffectHelperBracketOps[F[_], E](val self: Bracket[F, E]) extends AnyVal {
+  class EffectHelperBracketOps[F[_], E](val self: MonadCancel[F, E]) extends AnyVal {
```
x0ji0d, line 15: Not `BracketOps` anymore

rtar:
It is already updated to a newer version.

PR #103, opened 2021-09-07
Draft: migration to cats-effect 3

Z1kkurat:


core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
```diff
@@ -68,16 +68,13 @@ object CatsHelper {
...
+    def startEnsure(implicit F: Concurrent[F]): F[Fiber[F, Throwable, A]] = F.startEnsure(self)
 
 
     def toTry(implicit F: ToTry[F]): Try[A] = F.apply(self)
 
 
     def toFuture(implicit F: ToFuture[F]): Future[A] = F.apply(self)
-
-
-    def toResource(implicit F: Applicative[F]): Resource[F, A] = Resource.liftF(self)
```
t3hnar, line 80: you can deprecate, but not delete :) please

core/src/main/scala/com/evolutiongaming/catshelper/FeatureToggled.scala
```diff
@@ -75,7 +75,7 @@ object FeatureToggled {
 
       featureToggledResource = {
         val toggle = (v: Boolean) => toggleStateOf(v).flatMap { ts =>
-          flagRef.getAndSet(ts).flatMap(_.next.complete(ts)).uncancelable
+          Concurrent[F].uncancelable(_ => flagRef.getAndSet(ts).flatMap(_.next.complete(ts).void))
```
t3hnar, line 78: could you explain more about this construction?

core/src/main/scala/com/evolutiongaming/catshelper/Serial.scala
```diff
@@ -45,10 +46,10 @@ object Serial {
                 .void
             }
 
-            Concurrent[F].uncancelable {
+            Concurrent[F].uncancelable { _ =>
               for {
-                d <- Deferred.uncancelable[F, Either[Throwable, A]]
```
t3hnar, line 50: we might need to create own `uncancelable` it depends on usages.

core/src/main/scala/com/evolutiongaming/catshelper/TimerHelper.scala
```diff
@@ -1,21 +1,59 @@
...
-  implicit class TimerObjOpsTimerHelper(val self: Timer.type) extends AnyVal {
+  implicit class TimerObjOpsTimerHelper(val self: Temporal.type) extends AnyVal {
 
-    def empty[F[_] : Applicative]: Timer[F] = new Timer[F] {
+    def empty[F[_]](implicit F: Concurrent[F]): Temporal[F] = new Temporal[F] {
 
       val clock = Clock.empty[F]
 
-      def sleep(duration: FiniteDuration) = ().pure[F]
+      override def pure[A](x: A): F[A] = F.pure(x)
```
t3hnar, line 22: minor: `override` keywords are unnecessary here

core/src/test/scala/com/evolutiongaming/catshelper/TestIORuntime.scala
```diff
@@ -0,0 +1,7 @@
+package com.evolutiongaming.catshelper
+
+import cats.effect.unsafe.IORuntime
+
+trait TestIORuntime {
```
t3hnar, line 5: why do we need this trait at all? writing single line of 

```
implicit val ioRuntime: IORuntime = cats.effect.unsafe.IORuntime.global
```
is more or less equivalent to 

```
extends TestIORuntime
```

I'd personally prefer firs option.

core/src/test/scala/com/evolutiongaming/catshelper/ToFutureSpec.scala
```diff
@@ -25,6 +28,8 @@ class ToFutureSpec extends AsyncFunSuite with Matchers {
 
       val future = either.toFuture
 
+      Await.ready(future, 10.seconds)
```
t3hnar, line 31: same problem as with `ToTry`, we should then create custom implementation of `ToFuture`…

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/FeatureToggled.scala
Z1kkurat: it's basically the same, just changed with the introduction of Poll:  
https://typelevel.org/cats-effect/docs/migration-guide#bracket

core/src/main/scala/com/evolutiongaming/catshelper/Serial.scala
Z1kkurat: As I was told by TL folks at discord channel, their advice is to use `deferred.get.uncancelable` explicitly in the usage-site. I think that makes sense, will rework it here.

core/src/test/scala/com/evolutiongaming/catshelper/TestIORuntime.scala
Z1kkurat: I agree, this is a result of multiple steps of tinkering with test code, will remove that

core/src/main/scala/com/evolutiongaming/catshelper/FeatureToggled.scala
Z1kkurat: Actually, there's a syntax in MonadCancel extension, will use that

core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
Z1kkurat: I removed it because there's an identical method in `cats.effect.kernel.syntax.EffectResourceOps` which clashes with our own, so there's no need for that when using CE3

core/src/main/scala/com/evolutiongaming/catshelper/FeatureToggled.scala
Z1kkurat: done

core/src/main/scala/com/evolutiongaming/catshelper/Serial.scala
Z1kkurat: done

core/src/main/scala/com/evolutiongaming/catshelper/TimerHelper.scala
Z1kkurat: done

core/src/test/scala/com/evolutiongaming/catshelper/TestIORuntime.scala
Z1kkurat: done

core/src/test/scala/com/evolutiongaming/catshelper/ToFutureSpec.scala
Z1kkurat: Done. Please check the implementation. I think it _should_ be correct.

FunFunFine (approved):

core/src/main/scala/com/evolutiongaming/catshelper/Serial.scala
```diff
@@ -45,17 +46,17 @@ object Serial {
...
+                d <- Deferred.apply[F, Either[Throwable, A]]
+                f  = fa.attempt.flatMap { a => d.complete(a).void }
                 r <- ref.modify {
                   case Some(fs) => ((f :: fs).some, void)
                   case None     => (List.empty[F[Unit]].some, start(f))
                 }
                 _ <- r
               } yield for {
-                a <- d.get
+                a <- d.get.uncancelable
```
t3hnar, line 59: I think we don't need this anymore

DeviLab: done

core/src/main/scala/com/evolutiongaming/catshelper/EffectHelper.scala
```diff
@@ -107,16 +107,16 @@ object EffectHelper {
 
   implicit class FOpsEffectHelper[F[_], A](val self: F[A]) extends AnyVal {
 
-    def redeem[B, E](recover: E => B, map: A => B)(implicit bracket: Bracket[F, E]): F[B] = {
+    def redeem[B, E](recover: E => B, map: A => B)(implicit bracket: MonadCancel[F, E]): F[B] = {
       bracket.redeem(self)(recover, map)
     }
 
-    def redeemWith[B, E](recover: E => F[B], flatMap: A => F[B])(implicit bracket: Bracket[F, E]): F[B] = {
+    def redeemWith[B, E](recover: E => F[B], flatMap: A => F[B])(implicit bracket: MonadCancel[F, E]): F[B] = {
```
t3hnar, line 114: could you please rename everywhere

```implicit bracket: MonadCancel```
to 

```implicit monadCancel: MonadCancel```

DeviLab: done

core/src/test/scala/com/evolutiongaming/catshelper/ToFutureSpec.scala
t3hnar: @DeviLab could you please remove all added `Awaits` in this MR, as it should work without them otherwise we have bugs…

DeviLab: removed

t3hnar (approved):

PR #114, opened 2021-11-15
Implement Log on top of logback without slf4j

dfahritdinov:


core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -134,6 +134,38 @@ object Log {
...
+    import ch.qos.logback.classic.Level
+    import ch.qos.logback.classic.spi.LoggingEvent
+
+    import scala.collection.JavaConverters._
+
+    val FQCN = getClass.getName
+
+    def append(msg: => String, mdc: Mdc, level: Level, throwable: Throwable = null): F[Unit] = Sync[F].delay {
+      val event = new LoggingEvent(FQCN, logger, level, msg, throwable, null)
+      mdc.context.map(_.toSortedMap) foreach { mdc =>
```
t3hnar, line 148: could it be too expensive to call `toSortedMap` on every log call?

t3hnar: what if logging is disabled ?

core/src/test/scala/com/evolutiongaming/catshelper/LogSpec.scala
```diff
@@ -78,6 +78,16 @@ class LogSpec extends AnyFunSuite with Matchers {
 
     io.unsafeRunSync() shouldEqual null
   }
+
+  test("logback implementation") {
+    val io = for {
+      logOf <- LogOf.logback[IO]
+      log <- logOf(getClass)
+      _ <- log.info("hello from logback", Log.Mdc("k" -> "test value for K"))
```
t3hnar, line 86: what are we testing here? :)

build.sbt
```diff
@@ -51,8 +51,8 @@ lazy val core = project
       Cats.effect,
       machinist,
       `slf4j-api`,
+      logback,
```
t3hnar, line 54: so logbook is now required for all? would we have any problems with that?

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
dfahritdinov: Shouldn't be expensive as it's just cast: `asInstanceOf`

core/src/test/scala/com/evolutiongaming/catshelper/LogSpec.scala
dfahritdinov: possibility to log message and see it as output. Do you know better way to test such logger?

build.sbt
dfahritdinov: I don't expect any issues as we're already using logback as implementation for slf4j

core/src/test/scala/com/evolutiongaming/catshelper/LogSpec.scala
t3hnar: register custom appenders from code in test with vars and check those are assigned to correct values?

PR #115, opened 2021-11-16
Add convenient way to call FromFuture

FunFunFine:
Currently it's not possible to call apply on the classic 'summon' method:
```scala
val future: Future[Unit] = ???

def example[F[_]: FromFuture] = {
  val fa0 = FromFuture[F[_]](future) //fails to compile
  val fa1 = FromFuture.apply[F](future) //fails to compile
  val fa2 = FromFuture[F].apply(future) //compiles, but triggers you to replace apply with application
  val fa3 = FromFuture.summon[F](future) //fails to compile
  val fa4 = FromFuture.summon[F].apply(future) //compiles, but ugly
}
```
And having `FromFuture` instance as an explicit parameter is not always convenient. 

This PR introduces new partially-applied method to call this like that:

```scala
val fa = FromFuture.defer[F](future)
```

Note that it is possible to achieve the `fa0` way of calling it just by renaming new `defer` method to apply, but this will break existing `apply` method, which summons the instance. Perhaps it's possible to deprecate it in favour of `summon` method, and then after its removal introduce said `apply` method.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
@@ -24,7 +24,7 @@ object FromFuture {
   def defer[F[_]]: FromFutureDeferPA[F] = new FromFutureDeferPA[F]()
 
   class FromFutureDeferPA[F[_]](val __ : Boolean = true) extends AnyVal {
-    def apply[A](future: => Future[A])(implicit instance: FromFuture[F]): F[A] = instance(future)
+    def apply[A](future: => Future[A])(implicit F: FromFuture[F]): F[A] = F(future)
   }
 
   def apply[F[_]](implicit F: FromFuture[F]): FromFuture[F] = F
--- a/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
@@ -14,12 +14,25 @@ trait FromFuture[F[_]] {
 
 object FromFuture {
 
+  /** Convinient summon-and-run function
+   *
+   * @example {{{
+   *          class Foo[F[_]: FromFuture: Monad](db: FutureDB){
+   *            def query(lines: Int): F[Unit] = FromFuture.defer[F](db.query(lines))
+   *          }
+   * }}} */
+  def defer[F[_]]: FromFutureDeferPA[F] = new FromFutureDeferPA[F]()
+
+  class FromFutureDeferPA[F[_]](val __ : Boolean = true) extends AnyVal {
+    def apply[A](future: => Future[A])(implicit instance: FromFuture[F]): F[A] = instance(future)
+  }
+
   def apply[F[_]](implicit F: FromFuture[F]): FromFuture[F] = F
 
   def summon[F[_]](implicit F: FromFuture[F]): FromFuture[F] = F
 
 
-  implicit def lift[F[_]: Async](implicit executor: ExecutionContext): FromFuture[F] = {
+  implicit def lift[F[_] : Async](implicit executor: ExecutionContext): FromFuture[F] = {
 
     new FromFuture[F] {
 
@@ -42,7 +55,7 @@ object FromFuture {
   }
 
 
-  def functionK[F[_]: FromFuture]: FunctionK[Future, F] = new FunctionK[Future, F] {
+  def functionK[F[_] : FromFuture]: FunctionK[Future, F] = new FunctionK[Future, F] {
 
     def apply[A](fa: Future[A]) = FromFuture.summon[F].apply(fa)
   }
--- a/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
@@ -23,7 +23,7 @@ object FromFuture {
    * }}} */
   def defer[F[_]]: FromFutureDeferPA[F] = new FromFutureDeferPA[F]()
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/evolutiongaming/catshelper/SerialKey.scala
```diff
@@ -17,7 +17,7 @@ object SerialKey {
     def apply[A](key: K)(task: F[A]) = task.map { _.pure[F] }
   }
 
-  def of[F[_]: Concurrent: Parallel: Runtime, K: Hash]: F[SerialKey[F, K]] = {
```
t3hnar, line 20: these are breaking changes and not relevant to original problem, could you please deprecate this method and create a new one as separate MR? as I can accept other part already :)

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
```diff
@@ -14,12 +14,25 @@ trait FromFuture[F[_]] {
...
 object FromFuture {
 
+  /** Convinient summon-and-run function
+   *
+   * @example {{{
+   *          class Foo[F[_]: FromFuture: Monad](db: FutureDB){
+   *            def query(lines: Int): F[Unit] = FromFuture.defer[F](db.query(lines))
+   *          }
+   * }}} */
+  def defer[F[_]]: FromFutureDeferPA[F] = new FromFutureDeferPA[F]()
```
t3hnar, line 24: this is nice!

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
```diff
@@ -14,12 +14,25 @@ trait FromFuture[F[_]] {
...
+
+  class FromFutureDeferPA[F[_]](val __ : Boolean = true) extends AnyVal {
```
t3hnar, line 26: private ?

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
```diff
@@ -14,12 +14,25 @@ trait FromFuture[F[_]] {
...
+    def apply[A](future: => Future[A])(implicit instance: FromFuture[F]): F[A] = instance(future)
```
t3hnar, line 27: usually we call this `F`, just an advice to rename `instance` to `F`, not a review blocker though

core/src/test/scala/com/evolutiongaming/catshelper/SerialKeyTest.scala
```diff
@@ -378,7 +378,7 @@ object SerialKeyTest {
 
   object Queue {
 
-    def of[F[_]: Concurrent: Parallel, K: Hash, A]: F[Queue[F, K, A]] = {
+    def of[F[_]: Concurrent, K: Hash, A]: F[Queue[F, K, A]] = {
```
t3hnar, line 381: these are breaking changes and not relevant to original problem, could you please deprecate this method and create a new one as separate MR? as I can accept other part already :)

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/SerialKey.scala
FunFunFine: Sure, I just no-brain fixed some warnings :)

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
FunFunFine: I'm afraid it's not possible as `class FromFutureDeferPA escapes its defining scope blah blah...`

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -76,7 +78,92 @@ object Log {
 
   def summon[F[_]](implicit F: Log[F]): Log[F] = F
 
-  def apply[F[_]: Sync](logger: Logger): Log[F] = new Log[F] {
+  def cached[F[_] : Sync](source: String, factory: ILoggerFactory): Log[F] = new Log[F] {
```
t3hnar, line 81: it is messed up a bit now - logger related stuff should not be part of this MR

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -3,7 +3,7 @@ package com.evolutiongaming.catshelper
 import cats.data.NonEmptyMap
 import cats.effect.Sync
```
FunFunFine, line 4: Wrong files, I'll remove them from this MR

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
t3hnar: try `private[FromFuture]`

FunFunFine: This would mean that it would be only accessible from object `FromFuture` I believe.
But when user writes `val d = FromFuture.defer[F]`, the type of `d` will be `FromFutureDeferPA`,and that won't work.

t3hnar: just try :)

FunFunFine: Wow, that's something new for me, thanks.

t3hnar:
released under `v2.7.0`

Issue #117, opened 2021-11-28
Imports unification and other syntax changes 


FunFunFine:
There is a little incosistency regarding imports in the library: while any other library (e.g. cats, cats-effect & tofu) import their syntax by `import lib.syntax.something._` and `import lib.syntax.all._`, `cats-helper` does it in its own unique way by `import com.evolutiongaming.SomethingHelper._`. 

It kind of slows down the adoption by newcomers, because they can easily miss that those syntax objects even exist, so I would like to propose to change this import scheme to be the same as in cats. WDYT?

Other thing I would like to discuss here (I don't think it is worth to create separate issue for that) is the pollution of `apply` methods. One can easily find such code in `cats-helper` (e.g. Blocking, [FromFuture](#115)):
```scala
trait Something[F[_]] {
  def apply(something: Thing): F[OtherThing]
}

object Something {
  def apply[F[_]: Something]: Something[F] = implicitly
}
```
Not only that it just doesn't work to call it like `Something[F](thing)` because of double apply and thing being treated as explicitly passed instance, but IntelliJ IDEA often fails to find usages or asks if I want to go to instance or to method. In other words it is inconvenient.  So my proposal is to add alternative methods names with possible deprecation of `apply` methods. 

I could get this done, but I would like to see some discussion here. Are those proposals any good?

t3hnar:
@FunFunFine not sure this does really makes a lot of sense, as comparing to cats/cats-effect - we don't have much to import hence not sure feature of being able to "selectively import something" is useful

As drawback - there is added maintenance and runtime cost of flexible solution

FunFunFine:
> @FunFunFine not sure this does really makes a lot of sense, as comparing to cats/cats-effect - we don't have much to import hence not sure feature of being able to "selectively import something" is useful
> 
> As drawback - there is added maintenance and runtime cost of flexible solution

My proposal is more about `syntax.all._` part. Currently there are seven different `*Helper` objects which need to be imported separately. Maybe we should allow them all to be imported with just one object then (e.g. `import com.evolutiongaming.catshelper.syntax._`)?

mr-git:
With upgrade to CE3 and changing package to `com.evolution`, this would be very welcome change (if one is afraid to add such nicer import in point release).

I've had numerous times, when I had to lookup which is the required import, thus having something common, very similar to Cats and Cats-Effect approach, would be very welcome usability improvement.

Nobody is going to remove the ability to have explicit single given imports.

PR #119, opened 2021-12-01
Refactor syntax objects due to #117

FunFunFine:
I've seen some support in #117, so I made this PR.
It does several things:
- introduces `syntax` object and cats-style val's to import syntax;
- reworks `*Helper` objects to be traits & object in order to create `syntax.all`
- creates `com.evolutiongamin.syntax.all` objects.

Note: this is a breaking change, because someone could have been importing implicit classes explicitly, but now they are not available.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
@@ -11,8 +11,36 @@ import scala.concurrent.Future
 import scala.reflect.ClassTag
 import scala.util.{Either, Try}
 
-object CatsHelper {
+object CatsHelper extends CatsSyntax
 
+trait CatsSyntax {
+
+  import CatsSyntaxOps._
+
+  implicit def toApplicativeErrorOpsCatsHelper[F[_], E](self: ApplicativeError[F, E]): ApplicativeErrorOpsCatsHelper[F, E] = new ApplicativeErrorOpsCatsHelper(self)
+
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
```diff
@@ -45,20 +73,24 @@ object CatsHelper {
 
       for {
         started <- Deferred[F, Unit]
-        fiber   <- faOf(started).start
-        _       <- started.get
+        fiber <- faOf(started).start
+        _ <- started.get
```
t3hnar, line 77: not sure this is nicer now :)

core/src/main/scala/com/evolutiongaming/catshelper/ParallelHelper.scala
```diff
@@ -2,11 +2,26 @@ package com.evolutiongaming.catshelper
 
 import cats.{Foldable, Monoid, Parallel}
 
-object ParallelHelper {
+object ParallelHelper extends ParallelSyntax
+
+trait ParallelSyntax {
```
t3hnar, line 7: why not move all `*Syntax` classes into own files?

t3hnar: also I propose to make alternative  `object *Syntax` and leave `Helper` for for backward compatibility

t3hnar (changes requested):

PR #131, opened 2022-01-12
add factory method for creating F[Log] from implicit LogOf instance

dfahritdinov:


core/src/main/scala/com/evolutiongaming/catshelper/LogOf.scala
```diff
@@ -19,6 +21,9 @@ object LogOf {
 
   def summon[F[_]](implicit F: LogOf[F]): LogOf[F] = F
 
+  def logger[F[_]: LogOf, C: ClassTag]: F[Log[F]] = apply[F].apply(implicitly[ClassTag[C]].runtimeClass)
```
t3hnar, line 24: could we cal it `log` after the class `Log` ?

t3hnar:
Nice idea!

core/src/main/scala/com/evolutiongaming/catshelper/LogOf.scala
dfahritdinov: done

core/src/main/scala/com/evolutiongaming/catshelper/LogOf.scala
```diff
@@ -21,9 +21,9 @@ object LogOf {
 
   def summon[F[_]](implicit F: LogOf[F]): LogOf[F] = F
 
-  def logger[F[_]: LogOf, C: ClassTag]: F[Log[F]] = apply[F].apply(implicitly[ClassTag[C]].runtimeClass)
+  def log[F[_]: LogOf, C: ClassTag]: F[Log[F]] = apply[F].apply(implicitly[ClassTag[C]].runtimeClass)
```
t3hnar, line 24: regarding this - I propose to add extension method to `Log` class, and reuse that method, so `ClassTag` functionality is available in both cases

dfahritdinov: done

t3hnar:
Thanks

PR #135, opened 2022-01-18
add ResourceManager

dfakhritdinov:


core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
```diff
@@ -0,0 +1,86 @@
...
+ *        l1 <- local()
+ *        l2 <- local()
+ *        l3 <- local()
+ *        // use allocated instances l1, l2, l3
+ *      } yield ()
+ *   }
+ *   // release l1, l2, l3
+ * }}}
+ */
+sealed trait ResourceManager[F[_]] {
```
t3hnar, line 28: why it is `sealed` ?

core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * ResourceManager abstracts dynamically allocating resources in scope of one [[Resource]]
+ *
+ * Example:
+ * {{{
+ *   def local(): Resource[IO, Unit] = ???
+ *
+ *   val global = ResourceManager[IO]
+ *
+ *   val io = global.use { manager =>
```
t3hnar, line 17: looks like example is not complete, there is no call of `register`

core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
```diff
@@ -0,0 +1,86 @@
...
+sealed trait ResourceManager[F[_]] {
+
+  def register[A](resource: Resource[F, A]): F[A]
+
+}
+
+object ResourceManager {
+
+  /** Raised if [[ResourceManager]] used outside of it's Resource scope */
+  final case object ResourceManagerAlreadyReleasedException extends RuntimeException with NoStackTrace
```
t3hnar, line 37: please rename
```
ResourceManager.ResourceManagerAlreadyReleasedException
```
to

```
ResourceManager.AlreadyReleasedException
```

core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
t3hnar: minor: I'd also suggest to rename it to `ResourceRegistry`

core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
```diff
@@ -0,0 +1,86 @@
...
+        new ResourceManager[F] {
+
+          /**
+           * Allocate [[resource]] and remember it's release effect if [[ResourceManager]] still in running state,
+           * otherwise release the [[resource]] immediately and raise exception
+           */
+          def register[A](resource: Resource[F, A]): F[A] = {
+            resource.allocated.flatMap {
+              case (allocation, release) =>
+                state.modify {
```
t3hnar, line 71:
```scala
state.modify {
                   case running: State.Running[F] => running + release -> allocation.pure[F]
                   case _                         => State.Released[F]() -> release.attempt *> ResourceManagerAlreadyReleasedException.raiseError
                 }.flatten

```

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/ResourceManager.scala
dfahritdinov: why so? That makes code more complicated

t3hnar: just a suggestion :)
imho that does make code simpler because you don't have additional `Boolean` variables, but I'm fine even with prev version.

PR #140, opened 2022-02-01
Log logback: check log level before appending message

dfahritdinov:


core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -144,11 +144,13 @@ object Log {
     val FQCN = getClass.getName
 
     def append(msg: => String, mdc: Mdc, level: Level, throwable: Throwable = null): F[Unit] = Sync[F].delay {
-      val event = new LoggingEvent(FQCN, logger, level, msg, throwable, null)
-      mdc.context.map(_.toSortedMap) foreach { mdc =>
-        event.setMDCPropertyMap(mdc.asJava)
+      if (logger.isEnabledFor(level)) {
```
t3hnar, line 147: ouch

t3hnar (approved):

PR #152, opened 2022-04-28
remove ExecutorContext from FromFuture.lift

dfahritdinov:


olegkovalenko:
the change is breaking thus would require a major version bump

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
```diff
@@ -32,18 +32,19 @@ object FromFuture {
   def summon[F[_]](implicit F: FromFuture[F]): FromFuture[F] = F
 
 
-  implicit def lift[F[_] : Async](implicit executor: ExecutionContext): FromFuture[F] = {
```
t3hnar, line 35: backward incompatible change, could you please create a new method without ExecutionContext arguments aside reusing a previous?

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
dfahritdinov: Done. Unfortunately now new factory method cannot be `implicit` in the same scope

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
```diff
@@ -54,20 +53,41 @@ object FromFuture {
     }
   }
 
+  def asyncFromFuture[F[_]: Async]: FromFuture[F] = {
```
t3hnar, line 56: 1. I suggest to rename to `fromAsync`
2. removing `implicit` in the `lift` method is not a breaking change, so you still can make new `fromAsync` method an implicit :)

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/FromFuture.scala
dfahritdinov: Making the method implicit in the same scope as old `lift` will make them clash. Might be I should introduce new object like `Implicits` with new implicit method, but definitely not bring it on the same scope as old `lift`

t3hnar: but you should remove `implicit` keyword from old `lift`!

dfahritdinov: Ah, got it. I was thinking that removing `implicit` from old `lift` is braking change that you'd like to avoid.

PR #162, opened 2022-08-02
feat: add `combineWith` method to Mdc

kpodsiad:
Tiny amount of sugar for folks who like methods.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,12 @@ project/boot/
 project/plugins/project/
 .bsp
 
+# metals & bloop
+.bloop
+.metals
+metals.sbt
+.vscode
+
 # Scala-IDE specific
 .scala_dependencies
 .worksheet
--- a/core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
@@ -40,7 +40,9 @@ trait Log[F[_]] {
 
 object Log {
 
-  sealed trait Mdc
+  sealed trait Mdc {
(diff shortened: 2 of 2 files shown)


t3hnar:
@kpodsiad why implement `Semigroup[Mdc]` instead? :)

kpodsiad:
@t3hnar sorry, could you rephrase?

t3hnar:
> @t3hnar sorry, could you rephrase?

if you implement implicit instance of `Semigroup` for `Mdc` - then you will have all those helpers methods out of the box

```scala
object Mdc { 
  …
  implicit val semigroupMdc: Semigroup[Mdc] = ???
  …
}
```

t3hnar:
and I just checked - it has already `Semigroup` implemented, it should be `Monoid` though…

```scala
    implicit final val mdcSemigroup: Semigroup[Mdc] = Semigroup.instance {
      case (Empty, right) => right
      case (left, Empty) => left
      case (Context(v1), Context(v2)) => Context(v1 ++ v2)
    }    
```
implemented, then the question - why do you need this method at all? if `import cats.syntax.all._` should add proposed method? :)

kpodsiad:
Now I understand your original answer, but I also see that you answered it ;)

> why do you need this method at all? if import cats.syntax.all._ should add proposed method?

It's a simple matter of better user experience and leveraging method instead extension functions, which are easier to discover. Of course you can `import cats.syntax.all._` like you wrote but it pollutes namespace with, most of the time, useless methods unfortunately.

That being said, I've created this PR mostly because I was using the later approach and I wanted sth more convenient. However, through time my need to combine mdcs was gone and I don't need it anymore, so I don't want to advocate too much on it ;)

t3hnar:
ok :) 
I'm closing it then.

PR #170, opened 2022-09-19
libs update

BrainHorse:


project/Dependencies.scala
```diff
@@ -2,17 +2,17 @@ import sbt._
 
 object Dependencies {
 
-  val scalatest   = "org.scalatest"     %% "scalatest"       % "3.2.11"
-  val `slf4j-api` = "org.slf4j"          % "slf4j-api"       % "1.7.36"
-  val logback     = "ch.qos.logback"     % "logback-classic" % "1.2.11"
+  val scalatest   = "org.scalatest"     %% "scalatest"       % "3.2.13"
+  val `slf4j-api` = "org.slf4j"          % "slf4j-api"       % "2.0.1"
```
t3hnar, line 6: no breaking changes there?

BrainHorse: No, there are no, according to [documentation](https://www.slf4j.org/manual.html#fluent):

> The the 2.0.x series requires Java 8 and adds a backward-compatible fluent logging api.

t3hnar:
Thanks

project/Dependencies.scala
kpodsiad: it costed me almost a day to trace why our logging stopped working properly almost 2w ago, so I guess there were some breaking changes afterall :D

what's funny, everything was working locally

t3hnar: WOW!

@kpodsiad have you managed to trace it further? why it stopped to work?

kpodsiad: Yes, I was able to track where issue started and connect it to the MR with cats-helper bump.

We were using [scribe for sl4j 1.x](https://github.com/outr/scribe/blob/master/slf4j/src/main/scala/org/slf4j/impl/StaticLoggerBinder.scala) and simply using [scribe for sl4j 2.x](https://github.com/outr/scribe/blob/master/slf4j2/src/main/scala/scribe/slf4j/ScribeServiceProvider.scala) helped. Provided links show the difference on how library bind with sl4j and this changed between 1.x and 2.x.

PR #194, opened 2023-01-06
move logback based Log implementation to separate module

dfakhritdinov:


logback/src/main/scala/com/evolutiongaming/catshelper/Logback.scala
```diff
@@ -0,0 +1,80 @@
...
+  *
+  * Direct logback usage required to overcome limitations of SLF4J MDC API.
+  * SLF4J MDC API heavily rely on [[ThreadLocal]], example: ch.qos.logback.classic.util.LogbackMDCAdapter
+  *
+  * Logback' [[LoggingEvent]] allow setting MDC directly as Java map that should have performance benefits compared with SLF4J implementation.
+  *
+  * Please be aware that defining other version of logback might bring '''RUNTIME ERRORS''' or '''MISSING LOGS''' in case of binary incompatibility between them.
+  * Suggested approach is in using exactly same logback version as used in `cats-helper-logback` (among all others available through transitive dependencies)
+  */
+object Logback {
```
t3hnar, line 22: could we call it `LogOfFromLogback` ?

t3hnar (changes requested):

logback/src/main/scala/com/evolutiongaming/catshelper/Logback.scala
dfakhritdinov: done

t3hnar (approved):

PR #224, opened 2023-06-12
Add ResourceHelper

gjojgov-evolution:


core/src/main/scala/com/evolutiongaming/catshelper/ResourceHelper.scala
```diff
@@ -0,0 +1,47 @@
+package com.evolutiongaming.catshelper
+
+import cats.implicits._
+import cats.effect.{Resource, Sync}
+
+object ResourceHelper {
```
t3hnar, line 6: why not put this under `CatsHelper` ?

core/src/main/scala/com/evolutiongaming/catshelper/ResourceHelper.scala
```diff
@@ -0,0 +1,47 @@
...
+        timedRelease = for {
+          getMeasurement <- MeasureDuration[F].start
+          _              <- Log[F].info(s"$name releasing")
+          _ <- release.attemptTap {
+            case Left(err) => for {
+              measureResult <- getMeasurement
+              _ <- Log[F].error(s"$name release failed in ${measureResult.toMillis}ms with $err", err)
+            } yield ()
+
+            case Right(_) => for {
```
t3hnar, line 37: we un-match `Right(a)` and use `a` in `yield` :)

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/ResourceHelper.scala
gjojgov-evolution: Not sure I understand the comment correctly, but have a look at the updated version.

core/src/main/scala/com/evolutiongaming/catshelper/ResourceHelper.scala
gjojgov-evolution: Yep, moved!

t3hnar (approved):

PR #229, opened 2023-07-24
implement ResourceCounter

dfakhritdinov:
Motivation:

`ResourceCounter` provides caching mechanic for `Resource` by allocating source only once and caching its value & release effect. Every next `ResourceCounter.resource` will reuse cached value while release will be executed on releasing last resource `ResourceCounter.resource`.

`ResourceCounter` was inspired by Rust [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) and would be used in [smetrics](https://github.com/evolution-gaming/smetrics/pull/271)

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/ResourceCounter.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/ResourceCounter.scala
@@ -1,0 +1,66 @@
+package com.evolutiongaming.catshelper
+
+import cats.effect.{Concurrent, Resource}
+import cats.syntax.all.*
+
+trait ResourceCounter[F[_], A] {
+  def resource: Resource[F, A]
+}
+
+object ResourceCounter {
+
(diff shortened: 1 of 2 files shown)


Z1kkurat (approved):
LGTM, thank you!

core/src/main/scala/com/evolutiongaming/catshelper/ResourceCounter.scala
```diff
@@ -0,0 +1,87 @@
...
+                val s: State = State.Allocated(a, r, c + 1)
+                (s -> a).pure[F]
+            }
+
+          val cleanup =
+            state.update {
+
+              case State.Released =>
+                val msg = "releasing non-acquired ResourceCounter"
+                new IllegalStateException(msg).raiseError[F, State]
```
t3hnar, line 70: I'd prefer to not have an error here and potentially exploding whole composed `release` chain here

PR #236, opened 2023-08-21
pass MDC as lazy eval param

dfakhritdinov:


DeviLab (approved):

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -23,19 +23,19 @@ trait Log[F[_]] {
 
   @inline def error(msg: => String, cause: Throwable): F[Unit] = error(msg, cause, mdc = Log.Mdc.empty)
 
-  def trace(msg: => String, mdc: Log.Mdc): F[Unit]
+  def trace(msg: => String, mdc: => Log.Mdc): F[Unit]
```
t3hnar, line 26: this does break compatibility, am I right?

PR #237, opened 2023-08-22
Generalize ResourceLogOps.log to support arbitrary callbacks

edubrovski:


core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
```diff
@@ -146,21 +147,34 @@ object CatsHelper {
     }
   }
 
-  implicit final class ResourceLogOps[F[_], A](val self: Resource[F, A]) extends AnyVal {
-    def log(name: String)(implicit F: Sync[F], log: Log[F], md: MeasureDuration[F]): Resource[F, A] = Resource {
+  implicit final class ResourceObservabilityOps[F[_], A](val self: Resource[F, A]) extends AnyVal {
```
t3hnar, line 150: this is not backward compatible change :(

core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
```diff
@@ -146,21 +147,34 @@ object CatsHelper {
...
+
+    /**
+     * Add callbacks to a resource. Use for things like metrics and logs.
+     * It's very important that these callbacks shouldn't fail,
+     * otherwise there can be a resource leak.
+     */
+    def observe(
```
t3hnar, line 157: are you sure that this is useful? :)

could you please describe your use case

t3hnar (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
edubrovski: I think this is just how diff is displayed. I didn't rename ResourceLogOps, ResourceLogOps is still there.

core/src/main/scala/com/evolutiongaming/catshelper/CatsHelper.scala
edubrovski: This is to add metrics. Adding metrics to a resource is very similar to adding logs.  

see `consumerWithMetrics` in another PR:
https://github.com/evolution-gaming/kafka-journal/pull/500/files#diff-fdca83b54e7278ff458b068b2b38db53c143ce7138c3074defce921a321ce425R40

`consumerWithMetrics` looks pretty ugly, I want to avoid that

PR #245, opened 2023-10-06
Revert binary incompatible commit (ce2)

DeviLab:


dfakhritdinov:
Would it be possible to revert only my commit please? The one that brings lazy MDC params?

PR #246, opened 2023-10-09
Implement lazy MDC

DeviLab:


version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "3.8.1-SNAPSHOT"
+ThisBuild / version := "3.9.0-SNAPSHOT"
```
dfakhritdinov, line 1: you should not bump version manually, please do it as part of `sbt release`

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -44,30 +44,66 @@ object Log {
...
+      }
+    }
 
-    def fromMap(map: Map[String, String]): Mdc = fromSeq(map.toSeq)
+    val empty: Mdc = Empty
+
+    type Record = (String, String)
+
+    @deprecated("Use Mdc.Lazy.apply. If it's not enough - Mdc.Lazy.fromMap", "3.9.0")
+    def apply(head: Record, tail: Record*): Mdc = Lazy.fromMap( (head +: tail).toMap)
```
dfakhritdinov, line 65: Please do not deprecate those factory methods, in many places its fine to have eager construction

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -44,30 +44,66 @@ object Log {
   object Mdc {
 
     private object Empty extends Mdc
-    private final case class Context(values: NonEmptyMap[String, String]) extends Mdc {
-      override def toString: String = s"MDC(${values.toSortedMap.mkString(", ")})"
-    }
+    private final case class Context(getValues: () => Map[String, String]) extends Mdc {
```
dfakhritdinov, line 47: Might be better add new lazy implementation and keep eager one as is?

dfakhritdinov (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
DeviLab: Something like Mdc("key" -> parseSomethingFromJson(...)) can be heavy and would be nice to be lazy

version.sbt
DeviLab: It will offer "3.9.0" and I'll just accept it

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
DeviLab: When Mdc.fromMap/fromList is called, the parameters will be evaluated anyway, so it's actually eager (but implemented a bit differently)

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
dfakhritdinov: yes, but its up to user to choose eager/lazy to use

dfakhritdinov (changes requested):

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
DeviLab: When do you think the eager implementation is preferred and a user wants to choose it instead of the lazy one?

dfakhritdinov: 1. when MDC already pre-calculated 
2. when MDC is compact and all its values already available

I would say that almost everywhere I prefer eager. Lazy one is corner case in the apps with wide context

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
dfakhritdinov: additional reason of using eager implementation in eager factories is in avoiding `lazy val` where its definitely not needed (due to already pre-evaluated mdc map)

dfakhritdinov: Previous comment dropped by mistake:

> Then please elaborate why eager implementation should be done on top of lazy? I see no reason for it and prefer to keep both of them, but separately making the logic strait-forward and explicit

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
DeviLab: Discussed with @t3hnar and agreed to leave all these three methods as deprecated and implement another ones. Users will have to decide which one will be used

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
DeviLab: Discussed with @t3hnar and agreed to base lazy implementation on the eager one

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -44,30 +45,91 @@ object Log {
...
-      case (Empty, right) => right
-      case (left, Empty) => left
-      case (Context(v1), Context(v2)) => Context(v1 ++ v2)
+      def fromSeq(seq: Seq[Record]): Mdc = NonEmptyMap.fromMap(SortedMap(seq: _*)).fold(empty){ nem => EagerContext(nem) }
+
+      def fromMap(map: Map[String, String]): Mdc = fromSeq(map.toSeq)
+    }
+
+    object Lazy {
+      def apply(v1: => Record): Mdc = LazyContext(Eager(v1))
```
t3hnar, line 88: why these specialized functions are part of `Lazy` and not just `Mdc` ?

dfakhritdinov (approved):

PR #250, opened 2023-11-29
implement Log.withMdc for creating Log with MDC preset

dfakhritdinov:


mikost:
On line 144 in the changed file, `Log.scala`, there was already a similar, albeit slightly different, method: 
```
    def withMDC(mdc: Log.Mdc)(log: => Unit): Unit = {
      import Mdc.MdcOps
      mdc.context match {
        case None => log
        case Some(mdc) =>
          val backup = MDC.getCopyOfContextMap
          MDC.clear()
          mdc.toSortedMap foreach { case (k, v) => MDC.put(k, v) }
          log
          if (backup == null) MDC.clear() else MDC.setContextMap(backup)
      }
    }
```
I can't immediately discern whether that's for the same use case (as catered for in this MR) or another one. Can you elaborate, please? Perhaps some ScalaDoc-ing is warranted for these two? Or one of them should be deprecated?

PR #317, opened 2025-01-07
Add documentation for `Log` and `LogOf`

rtar:


core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
```diff
@@ -11,6 +11,18 @@ import org.slf4j.{Logger, MDC}
...
 import scala.collection.immutable.SortedMap
 
+/** Context specific logger instance.
+  * 
+  * Use [[LogOf]] to create the new instances of the class.
+  * 
+  * The recommendation is to avoid passing `Log` instances implicitly as there
+  * could be multiple instances of `Log`, which could lead to confusion and
+  * log messages attributed to the wrong class, which leaked its own `Log`
+  * instances accidentally.
```
mr-git, line 21: I am not insisting, but maybe it is worth to change the wording to NOT use negation as first option? Something like:
```suggestion
  * It is recommended to create new `Log` instance per class.
  * 
  * One can create instance per feature and pass it around implicitly, but
  * it can lead to confusion during attempt to find the source location 
  * of logged message.
  *
  * Avoid passing `Log` instances implicitly as there could be multiple 
  * instances of `Log` in scope.
```

rtar: I tried to play with the wording of the statement, but could not write anything better.

I think the whole idea of this documentation is to avoid passing `Log` implicitly, hence negative statement.

But I am fine to change it if you feel passing implicitly is okay in this case.

mr-git (approved):

core/src/main/scala/com/evolutiongaming/catshelper/LogOf.scala
```diff
@@ -8,6 +8,110 @@ import org.slf4j.{ILoggerFactory, LoggerFactory}
...
+  * 
+  * object UserService {
+  * 
+  *   def of[F[_]: LogOf]: F[UserService[F]] =
+  *     for {
+  *       log <- LogOf[F].forClass[UserService]
+  *       service = new UserService[F](log)
+  *     } yield service
+  * 
+  * }
```
sirmax, line 48: Consider swapping the order of `class` and `object` here. When reading from top to bottom it would make it easier to trace the sequence "request `LogOf`" > "use it to create `Log`" > "use `Log`".

```suggestion
  * object UserService {
  * 
  *   def of[F[_]: LogOf]: F[UserService[F]] =
  *     for {
  *       log <- LogOf[F].forClass[UserService]
  *       service = new UserService[F](log)
  *     } yield service
  * 
  * }
  * 
  * class UserService[F[_]: Monad](log: Log[F]) {
  * 
  *  def create(user: User): F[Unit] =
  *    for {
  *      _ <- log.info(s"Creating user...")
  *      _ <- ...
  *    } yield ()
  *   
  * }
```

core/src/main/scala/com/evolutiongaming/catshelper/Log.scala
sirmax: I don't mind the order of sentences here, but I would too recommend dropping the _"The recommendation is to"_ noise. And double "which" in one sentence is, probably, too much.
```suggestion
  * Avoid passing `Log` instances implicitly. Passing implicit instances around could "leak" a logger
  * into an unrelated code, and create a confusion with wrongly attributed log messages.
```
(this might need a pass of `scalafmt`).

PR #345, opened 2025-09-05
fix context initialization

BrainHorse:
Initialising the context twice causes the log appender to write duplicate log entries

### DIFF
--- a/logback/src/main/scala/com/evolutiongaming/catshelper/LogOfFromLogback.scala
+++ b/logback/src/main/scala/com/evolutiongaming/catshelper/LogOfFromLogback.scala
@@ -28,8 +28,10 @@ object LogOfFromLogback {
     Sync[F].delay {
       // see SLF4J compatibility Readme section
       val slf4jCtx = Try { org.slf4j.LoggerFactory.getILoggerFactory().asInstanceOf[ch.qos.logback.classic.LoggerContext] }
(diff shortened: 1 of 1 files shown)


logback/src/main/scala/com/evolutiongaming/catshelper/LogOfFromLogback.scala
```diff
@@ -28,8 +28,11 @@ object LogOfFromLogback {
     Sync[F].delay {
       // see SLF4J compatibility Readme section
       val slf4jCtx = Try { org.slf4j.LoggerFactory.getILoggerFactory().asInstanceOf[ch.qos.logback.classic.LoggerContext] }
-      val context  = slf4jCtx.getOrElse(new ch.qos.logback.classic.LoggerContext())
-      new ContextInitializer(context).autoConfig()
+      val context  = slf4jCtx.getOrElse {
+        val ctx = new ch.qos.logback.classic.LoggerContext()
+        new ContextInitializer(ctx).autoConfig()
```
kromanovs, line 33: Don't we need to call autoConfig() upon slf4jCtx if it already existed?

kromanovs: Yes, indeed the autoConfig() will be eventually called from `org.slf4j.LoggerFactory#getILoggerFactory` in case of Logback provider

PR #403, opened 2026-08-18
Fix GroupWithin timer, enqueue cost and release

BrainHorse:
`startTimer` runs the delay in a detached fiber and nothing cancels it. A batch closed by size, or
by release, leaves that fiber asleep for the rest of the delay. With a long delay and a fast fill
rate the fibers pile up.

`enqueue` calls `Nel.size` on every element, so filling a batch of n costs n squared over two. A
batch of 100000 took more than 10 seconds, and 0.56 after the fix.

`release` does not wait for delivery. If the delay timer already took the batch out of the `Ref`,
release sees an empty state and returns while the handler still runs in the timer fiber. On
shutdown that batch is lost.

The scaladoc is part of this branch. It states what release waits for, that batch order is not
guaranteed, and that none of the batching guarantees hold for `GroupWithin.empty` or for
`size <= 1`.

One spec is committed ignored. Batch order is not guaranteed because a fiber takes its batch out of
the `Ref` before it acquires the semaphore, and there is no async boundary in between. That needs
real parallelism to reproduce, so the spec cannot be made deterministic. It records the behaviour
instead of guarding it.

`GroupWithin.apply` declares its state ADT inside the method body. Scala 3 lifts those
into public JVM classes whose constructor takes the enclosing local object as `$outer`,
so no user code can name or build them, but MiMa still reports them as removed. ProblemFilters.exclude
filter in build.sbt covers it. Scala 2.13 does not report them.

## Bug Fixes

* Prevented overlapping batch deliveries.
* Ensured timers close only their associated batches.
* Prevented stale or cancelled timers from delivering batches.
* Ensured releases wait for timer-closed batches and in-flight deliveries.
* Improved propagation of handler errors during release.
* Improved delivery ordering across timer- and size-triggered batches.
* Improved enqueue performance and batching reliability under high-volume workloads.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import Dependencies.*
+import com.typesafe.tools.mima.core.*
 import sbtversionpolicy.Compatibility.BinaryCompatible
 
 def crossSettings[T](scalaVersion: String, if3: Seq[T], if2: Seq[T]) = {
@@ -38,6 +39,15 @@ lazy val commonSettings = Seq(
   scalacOptsFailOnWarn := Some(false),
 )
 
+lazy val groupWithinLocalStateFilters = Seq(
+  mimaBinaryIssueFilters ++= Seq(
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Empty$"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full$"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Stopped$"),
+  ),
+)
+
 val alias: Seq[sbt.Def.Setting[?]] =
   addCommandAlias("check", "all scalafmtCheckRepo versionPolicyCheck Compile/doc") ++
     addCommandAlias("fmt", "+scalafmtRepo") ++
@@ -61,6 +71,7 @@ lazy val root = project
 lazy val core = project
   .settings(
     commonSettings,
+    groupWithinLocalStateFilters,
     // formerly this was a top-level module and thus it retains the old name
     name := "cats-helper",
     libraryDependencies ++= Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -41,10 +41,7 @@ lazy val commonSettings = Seq(
 
 lazy val groupWithinLocalStateFilters = Seq(
   mimaBinaryIssueFilters ++= Seq(
-    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Empty$"),
-    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full"),
-    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full$"),
-    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Stopped$"),
+    ProblemFilters.exclude[Problem]("com.evolutiongaming.catshelper.GroupWithin#S#3#Full*"),
   ),
 )
 
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -17,8 +17,8 @@ import scala.concurrent.duration._
  * [[GroupWithin.Settings.delay]] passes after the first element of the batch, whichever happens
  * first.
  *
- * What is guaranteed around serialisation, cancellation and release depends on the implementation.
- * Only the batching one gives any of it, see `GroupWithin.apply` and `GroupWithin.empty`.
+ * The guarantees around serialisation, cancellation and release depend on the implementation. Only
+ * the batching one gives them, see [[GroupWithin.apply]] and [[GroupWithin.empty]].
  *
  * {{{
  * GroupWithin[IO]
@@ -43,6 +43,8 @@ trait GroupWithin[F[_]] {
 object GroupWithin {
 
   /**
+   * A batch is closed by whichever limit is reached first, it does not wait for both.
+   *
    * @param delay
    *   time to wait after the first element of a batch before the batch is closed
    * @param size
@@ -103,7 +105,7 @@ object GroupWithin {
 
           /**
            * @param size
-           *   number of elements in `as`, counted so that enqueue does not traverse the batch
+           *   always equal to `as.size`, counted so that enqueue does not traverse the batch
            * @param closed
            *   completed when the batch is closed by size or by release, which stops its timer, and
            *   used as the identity of the batch so that a timer only closes the batch it was
@@ -203,7 +205,8 @@ object GroupWithin {
                 .modify {
                   case State(batch: Batch.Full, inFlight) =>
                     (State(Batch.stopped, inFlight + 1), batch.closed.complete(()) *> consume(batch.as))
-                  case state => (state.copy(batch = Batch.stopped), void)
+                  case state =>
+                    (state.copy(batch = Batch.stopped), void)
                 }
                 .flatten
               // A batch the delay timer already took is delivered by its own fiber, so release
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -11,16 +11,47 @@ import com.evolutiongaming.catshelper.ClockHelper._
 
 import scala.concurrent.duration._
 
+/**
+ * Collects elements into batches and passes each batch to a handler.
+ *
+ * A batch is closed when it reaches [[GroupWithin.Settings.size]] elements, or when
+ * [[GroupWithin.Settings.delay]] passes after the first element of the batch, whichever happens
+ * first. A pending batch is also closed when the resource is released. Handler calls are
+ * serialised, so batches do not overlap, but the order of batches is not guaranteed.
+ *
+ * {{{
+ * GroupWithin[IO]
+ *   .apply[Int](GroupWithin.Settings(delay = 100.millis, size = 10)) { batch => store(batch) }
+ *   .use { enqueue => enqueue(1) *> enqueue(2) }
+ * }}}
+ */
 trait GroupWithin[F[_]] {
   import GroupWithin._
 
+  /**
+   * @param settings
+   *   batch size and delay
+   * @param f
+   *   handler for a closed batch, called serially
+   * @return
+   *   an [[GroupWithin.Enqueue]], releasing the resource flushes the pending batch
+   */
   def apply[A](settings: Settings)(f: Nel[A] => F[Unit]): Resource[F, Enqueue[F, A]]
 }
 
 object GroupWithin {
 
+  /**
+   * @param delay
+   *   time to wait after the first element of a batch before the batch is closed
+   * @param size
+   *   number of elements that closes a batch
+   */
   final case class Settings(delay: FiniteDuration, size: Int)
 
+  /**
+   * Does not batch, ignores the settings and calls the handler with one element at a time.
+   */
   def empty[F[_]]: GroupWithin[F] = new GroupWithin[F] {
 
     def apply[A](settings: Settings)(f: Nel[A] => F[Unit]): Resource[F, Enqueue[F, A]] = {
@@ -31,6 +62,13 @@ object GroupWithin {
     }
   }
 
+  /**
+   * With `size <= 1` or `delay <= 0` there is nothing to batch, so the handler is called directly
+   * per element and no state is allocated.
+   *
+   * While batching, enqueue is uncancelable, so an accepted element is always part of a batch.
+   * After release, enqueue silently discards the element.
+   */
   def apply[F[_]: Temporal]: GroupWithin[F] = {
 
     new GroupWithin[F] {
@@ -112,6 +150,9 @@ object GroupWithin {
     }
   }
 
+  /**
+   * Accepts one element into the current batch.
+   */
   trait Enqueue[F[_], A] {
 
     def apply(a: A): F[Unit]
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -173,7 +173,7 @@ object GroupWithin {
                     if (full.isFilled) ((S.empty, inFlight + 1), s.closed.complete(()) *> consume(full.as))
                     else ((full, inFlight), void)
                   case (S.Stopped, inFlight) =>
-                    (S.stopped, inFlight), void)
+                    ((S.stopped, inFlight), void)
                 }.flatten
               } yield ()
             }
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -128,8 +128,7 @@ object GroupWithin {
             ref <- Ref[F].of((S.empty, 0))
           } yield {
 
-            def consume(as: Nel[A]) = semaphore.permit
-              .use { _ => f(as.reverse) }
+            def consume(as: Nel[A]) = semaphore.permit.use { _ => f(as.reverse) }
               .guarantee {
                 ref
                   .modify { case (s, inFlight) =>
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -187,8 +187,10 @@ object GroupWithin {
                       val full = s.append(a)
                       if (full.isFilled) ((S.empty, inFlight + 1), s.closed.complete(()) *> consume(full.as))
                       else ((full, inFlight), void)
-                    case (S.Empty, inFlight) => ((S.empty, inFlight), openBatch(a))
-                    case (S.Stopped, inFlight) => ((S.stopped, inFlight), void)
+                    case (S.Empty, inFlight) =>
+                      ((S.empty, inFlight), openBatch(a))
+                    case (S.Stopped, inFlight) =>
+                      ((S.stopped, inFlight), void)
                   }
                     .flatten
                 }
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -2,7 +2,7 @@ package com.evolutiongaming.catshelper
 
 import cats.data.{NonEmptyList => Nel}
 import cats.effect.implicits._
-import cats.effect.kernel.Ref
+import cats.effect.kernel.{Deferred, Ref}
 import cats.effect.std.Semaphore
 import cats.effect.{Clock, Concurrent, Resource, Temporal}
 import cats.implicits._
@@ -82,11 +82,23 @@ object GroupWithin {
         object S {
           def empty: S = Empty
           def stopped: S = Stopped
-          def full(as: Nel[A], timestamp: Long): S = Full(as, timestamp)
+          def full(a: A, timestamp: Long, closed: Deferred[F, Unit]): S = Full(Nel.of(a), 1, timestamp, closed)
 
           case object Empty extends S
           case object Stopped extends S
-          final case class Full(as: Nel[A], timestamp: Long) extends S
+
+          /**
+           * @param size
+           *   number of elements in `as`, counted so that enqueue does not traverse the batch
+           * @param closed
+           *   completed when the batch is closed by size or by release, which stops its timer
+           */
+          final case class Full(
+            as: Nel[A],
+            size: Int,
+            timestamp: Long,
+            closed: Deferred[F, Unit],
+          ) extends S
         }
 
         if (settings.size <= 1 || settings.delay <= 0.millis) {
@@ -100,16 +112,19 @@ object GroupWithin {
 
             def consume(as: Nel[A]) = semaphore.permit.use { _ => f(as.reverse) }
 
-            def startTimer(timestamp: Long) = {
-              val result = for {
-                _ <- Temporal[F].sleep(settings.delay)
-                a <- ref.modify {
+            def startTimer(timestamp: Long, closed: Deferred[F, Unit]) = {
+              val expire = ref
+                .modify {
                   case s: S.Full if s.timestamp == timestamp => (S.empty, consume(s.as))
                   case s => (s, void)
                 }
-                a <- a
-              } yield a
-              result
+                .flatten
+              Temporal[F]
+                .race(Temporal[F].sleep(settings.delay), closed.get)
+                .flatMap {
+                  case Left(_) => expire
+                  case Right(_) => void
+                }
                 .start
                 .void
             }
@@ -119,24 +134,26 @@ object GroupWithin {
               def apply(a: A) = {
                 Concurrent[F].uncancelable { _ =>
                   for {
-                    t <- Clock[F].nanos
-                    a <- ref.modify {
+                    timestamp <- Clock[F].nanos
+                    closed <- Deferred[F, Unit]
+                    action <- ref.modify {
                       case s: S.Full =>
                         val as = a :: s.as
-                        if (as.size >= settings.size) (S.empty, consume(as))
-                        else (s.copy(as = as), void)
-                      case S.Empty => (S.full(Nel.of(a), t), startTimer(t))
+                        val size = s.size + 1
+                        if (size >= settings.size) (S.empty, s.closed.complete(()) *> consume(as))
+                        else (s.copy(as = as, size = size), void)
+                      case S.Empty => (S.full(a, timestamp, closed), startTimer(timestamp, closed))
                       case S.Stopped => (S.stopped, void)
                     }
-                    a <- a
-                  } yield a
+                    _ <- action
+                  } yield {}
                 }
               }
             }
 
             val release = ref
               .modify {
-                case s: S.Full => (S.stopped, consume(s.as))
+                case s: S.Full => (S.stopped, s.closed.complete(()) *> consume(s.as))
                 case _ => (S.stopped, void)
               }
               .flatten
--- a/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
@@ -105,8 +105,8 @@ object GroupWithin {
 
           /**
            * @param size
-           *   number of elements in `as`, kept in step by `append` so that enqueue does not
-           *   traverse the batch
+           *   number of elements in `as`, kept in step by `add` so that enqueue does not traverse
+           *   the batch
            * @param closed
            *   completed when the batch is closed by size or by release, which stops its timer. Also
            *   identifies the batch, so a timer only closes the batch it was started for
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -38,6 +39,15 @@ lazy val commonSettings = Seq(
...
 )
 
+lazy val groupWithinLocalStateFilters = Seq(
+  mimaBinaryIssueFilters ++= Seq(
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Empty$"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Full$"),
+    ProblemFilters.exclude[MissingClassProblem]("com.evolutiongaming.catshelper.GroupWithin$S$3$Stopped$"),
+  ),
+)
```
mr-git, line 46: Please create an issue to remove these after the release, ideally, please provide PR with explicit statement that it should be merged after release vX.Y.Z.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -2,25 +2,60 @@ package com.evolutiongaming.catshelper
...
 
+/**
+ * Collects elements into batches and passes each batch to a handler.
+ *
+ * A batch is closed when it reaches [[GroupWithin.Settings.size]] elements, or when
+ * [[GroupWithin.Settings.delay]] passes after the first element of the batch, whichever happens
+ * first.
+ *
+ * What is guaranteed around serialisation, cancellation and release depends on the implementation.
+ * Only the batching one gives any of it, see `GroupWithin.apply` and `GroupWithin.empty`.
```
mr-git: It is hard to read, what is ".. any of it.."? Maybe state what the "it" is, like is it "guarantee" or "<something else>"?

Please make references linkable by Scaladoc, like:
```suggestion
 * What is guaranteed around serialisation, cancellation and release depends on the implementation.
 * Only the batching one gives any of it, see [[GroupWithin.apply]] and [[GroupWithin.empty]].
```

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -2,25 +2,60 @@ package com.evolutiongaming.catshelper
...
 
 object GroupWithin {
 
+  /**
+   * @param delay
+   *   time to wait after the first element of a batch before the batch is closed
+   * @param size
+   *   number of elements that closes a batch
+   */
   final case class Settings(delay: FiniteDuration, size: Int)
```
mr-git, line 53: should we mention explicitly, that batch/group closes on first condition met, that it doesn't wait for both conditions?

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+          def empty: Batch = Empty
+          def stopped: Batch = Stopped
+          def full(a: A, closed: Deferred[F, Unit]): Batch = Full(Nel.of(a), 1, closed)
+
+          case object Empty extends Batch
+          case object Stopped extends Batch
+
+          /**
+           * @param size
+           *   number of elements in `as`, counted so that enqueue does not traverse the batch
```
mr-git: what will happen if `as.length != size`?

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+           * @param closed
+           *   completed when the batch is closed by size or by release, which stops its timer, and
```
mr-git: can batch close by grouping delay too?

mr-git: probably we should mention delay (from public API) instead of internal "closed" deferred, which seems to get done on timer/delay, if my guess is correct

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+              semaphore.permit.use { _ => f(as.reverse) }.guarantee(delivered)
+            }
 
-            def startTimer(timestamp: Long) = {
-              val result = for {
-                _ <- Temporal[F].sleep(settings.delay)
-                a <- ref.modify {
-                  case s: S.Full if s.timestamp == timestamp => (S.empty, consume(s.as))
-                  case s => (s, void)
+            def startTimer(closed: Deferred[F, Unit]) = {
```
mr-git, line 147: why no explanation like for others?

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
-              } yield a
-              result
+                .flatten
+              closed.get
+                .timeoutTo(settings.delay, expire)
                 .start
                 .void
             }
 
+            def append(a: A, state: State, batch: Batch.Full) = {
```
mr-git: why no explanation like for others?

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
-                case s: S.Full => (S.stopped, consume(s.as))
-                case _ => (S.stopped, void)
-              }
-              .flatten
+            val release = {
+              val stop = ref
+                .modify {
+                  case State(batch: Batch.Full, inFlight) =>
+                    (State(Batch.stopped, inFlight + 1), batch.closed.complete(()) *> consume(batch.as))
+                  case state => (state.copy(batch = Batch.stopped), void)
```
mr-git: for readability
```suggestion
                  case State(batch: Batch.Full, inFlight) =>
                    (State(Batch.stopped, inFlight + 1), batch.closed.complete(()) *> consume(batch.as))
                  case state =>
                    (state.copy(batch = Batch.stopped), void)
```

kromanovs (approved):

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -36,6 +39,127 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
...
+    } yield {
+      overlaps shouldEqual 0
+      delivered.sorted shouldEqual (1 to elements).toList
+    }
+
+    program.timeout(60.seconds).unsafeRunSync()
+  }
+
+  // Demonstrates that the order of batches is not guaranteed, as stated in the GroupWithin
+  // scaladoc. A fiber takes its batch out of the Ref and only then acquires the semaphore, and
```
mr-git: it is `Scaladoc`, like `Scala`, with capital letter

```suggestion
  // Scaladoc. A fiber takes its batch out of the Ref and only then acquires the semaphore, and
```

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -36,6 +39,127 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
...
+      .use { enqueue => enqueue(1) *> enqueue(2) }
+
+    val test = TestControl.execute(program).flatMap { control =>
+      for {
+        _ <- control.tick
+        outcome <- control.results
+        _ <- IO { outcome shouldEqual Outcome.succeeded[Id, Throwable, Unit](()).some }
+        nextInterval <- control.nextInterval
+      } yield {
+        nextInterval shouldEqual Duration.Zero
```
mr-git, line 181: I fail to understand how this means that `timer` is cancelled?

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -36,6 +39,127 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
...
+
+  "enqueue at a constant cost for a large batch" in {
+    val elements = 100000
+    val settings = GroupWithin.Settings(delay = 1.day, size = elements + 1)
+    val program = GroupWithin[IO]
+      .apply[Int](settings) { _ => IO.unit }
+      .use { enqueue => (1 to elements).toList.traverse_ { enqueue.apply } }
+
+    program.timeout(10.seconds).unsafeRunSync()
+  }
```
mr-git, line 200: where is any check or assertion?

do we only test that `100_000` elements can be enqueued in under 10 seconds?

mr-git (commented):
I didn't dive into modified code, yet, I only skimmed part of PR!

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: It cannot diverge, as size is set to 1 at construction and incremented in the one place that adds elements to the list, the class itself and internal state of GroupWithin is not accessible by the library user.

build.sbt
BrainHorse: This is a very unnecessary complication. I will remove it after the release without issue, and even if I don't do that right away, those filters are legitimate. As for Scala users, those classes are inaccessible, and the likelihood of there being a Java library that uses those classes directly is negligible.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Not sure I understand the point here.

> can batch close by grouping delay too?

Yes, batch can be closed by a time delay. This is the core logic of this class.

> probably we should mention delay (from public API) instead of internal "closed" deferred, which seems to get done on timer/delay, if my guess is correct

Deferred is an internal implementation detail to cancel the delay fiber, so I don't get how it is related to the delay from public API.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Comment density/amount is not a target I am aiming for. I prefer self-explanatory code and to comment only the puzzling parts what the code cannot express. 'AppendConses' adds an element and increments a counter, and 'startTimer' is one expression. The non-obvious part of the timer that both identifies the batch and stops the timer is documented on the 'closed' parameter, which is where a reader first encounters it. The existing comments are there because something is not visible locally, such as why openBatch re-reads the state and why release waits for consumes that it did not start.

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
BrainHorse: This is a part of `TestControl`. 
```scala
  /**
   * Produces the minimum time which must elapse for a fiber to become eligible for execution.
   * If fibers are currently eligible for execution, or if the program is entirely deadlocked,
   * the result will be `Duration.Zero`.
   */
  val nextInterval: IO[FiniteDuration]
```

Therefore, if the value of nextInterval is not 0 after execution, it means that TestControl has some fibres to execute in the future.

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
BrainHorse: > where is any check or assertion?

No, the fact that this number of elements can be enqueued in under 10 seconds is a sign of success in itself.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+
+          /**
+           * @param size
+           *   number of elements in `as`, counted so that enqueue does not traverse the batch
+           * @param closed
+           *   completed when the batch is closed by size or by release, which stops its timer, and
+           *   used as the identity of the batch so that a timer only closes the batch it was
+           *   started for
+           */
+          final case class Full(as: Nel[A], size: Int, closed: Deferred[F, Unit]) extends Batch
```
mr-git: honestly, just by looking at code here, I cannot understand what `Stopped` and `Full` represents, similarly for the similarly named functions above. 

from other discussions, I can guess that `def full` is the "internal constructor" and that's explain why function `full` takes single element, while `Full` wants `NonEmptyList`

after reading code below, I understood that `Full` represents something like `InProgress`, `Waiting`, `Open` (for additional elements, to be closed by either addition of more elements or after the `delay`) - the `Full` carries meaning that no more elements can be added, but code describes different behaviour

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Done

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Done

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Done

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
BrainHorse: Done

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+
         if (settings.size <= 1 || settings.delay <= 0.millis) {
           val enqueue: Enqueue[F, A] = a => f(Nel.of(a))
           Resource.pure(enqueue)
         } else {
           val result = for {
             semaphore <- Semaphore[F](1)
-            ref <- Ref[F].of(S.empty)
+            drained <- Deferred[F, Unit]
+            ref <- Ref[F].of(State(Batch.empty, 0))
```
mr-git: with all the explicit machinery above, I almost want to see `State.empty` too - just for sake of "expresivness" :) 

also, please name the variables with words, that describe "what it does/holds", not its type (`ref: Ref` doesn't explain much to humans)

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
           } yield {
 
-            def consume(as: Nel[A]) = semaphore.permit.use { _ => f(as.reverse) }
+            // The caller counts the batch in while taking it out of the state, so the count is
+            // dropped here, once the handler has returned.
```
mr-git: honestly, even after reading the code, I still do not get this comment

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +91,126 @@ object GroupWithin {
...
+              val stop = ref
+                .modify {
+                  case State(batch: Batch.Full, inFlight) =>
+                    (State(Batch.stopped, inFlight + 1), batch.closed.complete(()) *> consume(batch.as))
+                  case state => (state.copy(batch = Batch.stopped), void)
+                }
+                .flatten
+              // A batch the delay timer already took is delivered by its own fiber, so release
+              // waits for every consume that started before it, not only for its own.
+              val awaitDelivery = ref.get.flatMap { state => drained.get.whenA(state.inFlight > 0) }
```
mr-git: nitpicking about readability - it might depend on person, but for many "when X then do Y" reads easier than "do Y only when X"

```suggestion
              val awaitDelivery = ref.get.flatMap { state => Temporal[F].whenA(state.inFlight > 0) { drained.get } }
```

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +93,127 @@ object GroupWithin {
...
+            // path that opens one. The state is read again after the allocation, because another
+            // fiber can open the batch in between.
+            def openBatch(a: A) = {
+              for {
+                closed <- Deferred[F, Unit]
+                action <- ref.modify {
+                  case state @ State(Batch.Empty, _) =>
+                    (state.copy(batch = Batch.full(a, closed)), startTimer(closed))
+                  case state @ State(batch: Batch.Full, _) => append(a, state, batch)
+                  case state => (state, void)
```
mr-git: for consistent style:

```suggestion
                  case state @ State(Batch.Empty, _) =>
                    (state.copy(batch = Batch.full(a, closed)), startTimer(closed))
                  case state @ State(batch: Batch.Full, _) =>
                    append(a, state, batch)
                  case state =>
                    (state, void)
```

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +93,127 @@ object GroupWithin {
...
-          final case class Full(as: Nel[A], timestamp: Long) extends S
+        sealed trait Batch
+
+        object Batch {
+          def empty: Batch = Empty
+          def stopped: Batch = Stopped
+          def full(a: A, closed: Deferred[F, Unit]): Batch = Full(Nel.of(a), 1, closed)
+
+          case object Empty extends Batch
+          case object Stopped extends Batch
```
mr-git: I have a feeling, that these are the "signals" not for a `Batch`, but for a `State`, unless the `State.batch` is used for 2 "features" - to manifest both the State's current state and to track the lifespan of the aggregating `Batch` - complicated...

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
mr-git: or there is a deadlock between some fibers :)

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -36,6 +39,127 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
...
+      } yield {
+        nextInterval shouldEqual Duration.Zero
+      }
+    }
+
+    test.unsafeRunSync()
+  }
+
+  "enqueue at a constant cost for a large batch" in {
+    val elements = 100000
```
mr-git:
```suggestion
    val elements = 100_000
```

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
mr-git: feels sketchy - what if CI "pauses" the runtime during the test? it will show up as flaky, wouldn't it?

but, I don't see the other way how to test that we can `enqueue` a lot of elements in single batch in "reasonable" time 😞

mr-git: maybe test that times the enqueuing of 10k vs 100k elements? the latter should not take more than 11x of former, right? though it could become flaky to

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -39,69 +93,127 @@ object GroupWithin {
...
-              result
+                .flatten
+              closed.get
+                .timeoutTo(settings.delay, expire)
                 .start
                 .void
             }
 
+            def append(a: A, state: State, batch: Batch.Full) = {
+              val as = a :: batch.as
```
mr-git: why we prepend here and `.reverse` on Line 145? if it is due to performance reasons of how `NonEmptyList` is formed, let's add comment here.

other option could be the`NonEmptyChain` - it has O(1) prepend and append - there wouldn't be a need for `.reverse`, would be?

BrainHorse:
A note on how this PR is shaped, since several threads cover the same topic.
The scope is the three defects described: the uncancelled timer fibre, the linear size check on every enqueue and the fact that release does not wait for delivery. Nothing else is touched.
No renaming of existing identifiers, no refactoring of untouched code and no reformatting.
The implementation was deliberately rewritten to stay as close as possible to the current code. An earlier revision of this branch renamed the state ADT, extracted helpers, and reflowed several blocks. This resulted in a larger diff and a long tail of review comments about naming and style rather than the fixes themselves. I reverted all of it. The state type is S again, S.Empty, S.Stopped and S.Full keep their names, ref keeps its name, and the in-flight counter lives in the ref as a tuple rather than as a new named type to be consistent with surrounding code style.
Suggestions that improve the surrounding code are welcome as follow-up PRs. Including them here increases the diff and makes the three fixes harder to review.

cc @mr-git

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: Agreed that 'Full' reads as 'no more elements accepted', while the code uses it for an open, accumulating batch. This naming convention remains unchanged in the current code, this PR only adds fields to it. Renaming the ADT would affect every branch of the method and explode the amount of changed lines.
`def full` is the internal constructor and takes a single element because a batch always opens with exactly one.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: The `State` no longer exists in this revision, so `State.empty` does not apply.
On reference, that name is unchanged from the current code. Renaming it would require amending the code, which this PR does not cover, so it is out of scope here.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: The prepend and reverse functions remain unchanged from the current code, only the append function has been relocated to `S.Full`. The pair is deliberate: `prepend` is O(1) per element and `reverse` is one O(n) pass per batch, so filling a batch is linear overall.
`NonEmptyChain`, I think, would not remove that pass, `toNonEmptyList` is itself O(n) and Nel is in the public signature f: Nel[A] => F[Unit], so the conversion has to happen somewhere. The total cost presumably remains the same (to prove NonEmptyChain improvements benchmarks are required), plus a change to untouched code.

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -76,32 +159,46 @@ object GroupWithin {
...
+            def openBatch(element: A): F[Unit] = {
+              for {
+                closed <- Deferred[F, Unit]
+                _ <- ref.modify {
+                  case (S.Empty, inFlight) => ((S.full(element, closed), inFlight), startTimer(closed))
+                  case (s: S.Full, inFlight) =>
+                    val full = s.append(element)
+                    if (full.isFilled) ((S.empty, inFlight + 1), s.closed.complete(()) *> consume(full.as))
+                    else ((full, inFlight), void)
+                  case (S.Stopped, inFlight) => ((S.stopped, inFlight), void)
```
mr-git:
```suggestion
                  case (S.Empty, inFlight) =>
                    ((S.full(element, closed), inFlight), startTimer(closed))
                  case (s: S.Full, inFlight) =>
                    val full = s.append(element)
                    if (full.isFilled) ((S.empty, inFlight + 1), s.closed.complete(()) *> consume(full.as))
                    else ((full, inFlight), void)
                  case (S.Stopped, inFlight) =>
                    (S.stopped, inFlight), void)
```

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -76,32 +159,46 @@ object GroupWithin {
...
-                    }
-                    a <- a
-                  } yield a
+                  ref.modify {
+                    case (s: S.Full, inFlight) =>
+                      val full = s.append(a)
+                      if (full.isFilled) ((S.empty, inFlight + 1), s.closed.complete(()) *> consume(full.as))
+                      else ((full, inFlight), void)
+                    case (S.Empty, inFlight) => ((S.empty, inFlight), openBatch(a))
+                    case (S.Stopped, inFlight) => ((S.stopped, inFlight), void)
```
mr-git:
```suggestion
                    case (S.Empty, inFlight) =>
                      ((S.empty, inFlight), openBatch(a))
                    case (S.Stopped, inFlight) =>
                      ((S.stopped, inFlight), void)
```

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -36,6 +39,162 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
...
+    val elements = 2000
+
+    val program = for {
+      inFlight <- Ref[IO].of(0)
+      overlaps <- Ref[IO].of(0)
+      delivered <- Ref[IO].of(List.empty[Int])
+      handler = (batch: Nel[Int]) =>
+        for {
+          entered <- inFlight.updateAndGet { _ + 1 }
+          _ <- overlaps.update { _ + 1 }.whenA(entered > 1)
```
mr-git:
```suggestion
          _ <- IO.whenA(entered > 1) { overlaps.update { _ + 1 } }
```

mr-git (approved):
left 3 nitpicking style suggestions

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
```diff
@@ -44,11 +98,27 @@ object GroupWithin {
...
+           * @param closed
+           *   completed when the batch is closed by size or by release, which stops its timer. Also
+           *   identifies the batch, so a timer only closes the batch it was started for
+           */
+          final case class Full(
+            as: Nel[A],
+            size: Int,
+            closed: Deferred[F, Unit],
+          ) extends S {
+            def append(a: A): Full = Full(a :: as, size + 1, closed)
```
Z1kkurat: is it not "prepend" though?

build.sbt
BrainHorse: @mr-git This is the only unresolved thread remaining. Does this block the PR from being merged?

Z1kkurat (approved):

core/src/main/scala/com/evolutiongaming/catshelper/GroupWithin.scala
BrainHorse: From the internal implementation point of view, it is 'prepend', but from the client side, we 'append'. I can change it to 'add' to make it less confusing.

Z1kkurat: if the order doesn't matter then let's rename that please to let the reader have no assumptions about the order of insertion

BrainHorse: The order does matter because the client receives the reverted list upon completion of the batch. As an alternative, I can propose inlining this without the name, as it was before.

Z1kkurat: i'm fine with either inlining or renaming to 'add', up to you!

BrainHorse: Done in [rename append to add](https://github.com/evolution-gaming/cats-helper/pull/403/commits/3f5c8b79f73998571830ff041561e5a99c265fbb)

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
mr-git: nitpicking: I'd add here comment on why this check is enough

core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
mr-git: nitpicking: IMHO comment on `10.seconds` is due - Next Developer will have hard time to understand why `10.seconds` were picked - only option will be to lookup PR and read over discussions, not the trivial feat

PR #406, opened 2026-08-22
Fix O(n²) batch fill in GroupWithin

stasimus:
GroupWithin recomputed NonEmptyList.size on every enqueued element to check the batch threshold, making batch fill O(n²). Adds a running `count` field to `S.Full` for an O(1) check instead. No behavior change, covered by a new size = 3 test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed item grouping so batches consistently respect the configured size limit.
  * Improved handling of groups containing more than two items, ensuring items are emitted in complete, correctly sized batches.
  * Ensured grouping remains accurate when processing items incrementally.

* **Tests**
  * Added coverage for grouping six items into two batches of three.
  * Verified that larger batch sizes produce complete, non-empty groups.

### DIFF
--- a/core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
+++ b/core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
@@ -26,6 +26,11 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
     `collect until size reached`[IO]
   }
 
+  "collect until size reached, size > 2" in ioTest { env =>
+    import env._
+    `collect until size reached, size > 2`[IO]
+  }
+
   "collect until deadline reached" in ioTest { env =>
     import env._
     `collect until deadline reached`[IO]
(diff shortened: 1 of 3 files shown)


core/src/test/scala/com/evolutiongaming/catshelper/GroupWithinSpec.scala
```diff
@@ -73,6 +78,27 @@ class GroupWithinSpec extends AnyFreeSpec with Matchers {
     }
   }
 
+  private def `collect until size reached, size > 2`[F[_]: Temporal] = {
```
mr-git, line 81: looks useless! how is this different from `collect until size reached`?

stasimus: Reaching a threshold of 3 requires the counter to be persisted once (after element 2) and read back (at element 3), which is exactly the path currently untested

mr-git: do we need `collect until size reached` - which has exactly the same shape, only checks till `size = 2` (which is still larger than special case of `size = 1`!)

mr-git: N.B.: I am NOT opposing the code change - I am just questioning the usefulness of the extra unit-test!

mr-git:
BTW, @stasimus, changes in this PR conflict with similar changes in https://github.com/evolution-gaming/cats-helper/pull/403!

Issue #404, opened 2026-08-21
Serial, SerialKey and SerParQueue never advance after a task cancels


BrainHorse:
Cancelling a task permanently blocks the queue. Any task queued afterwards is lost, and any caller waiting on one will wait indefinitely.

All three share the same wrapper:

```scala
d <- Deferred[F, Either[Throwable, A]]
task = task0.attempt.flatMap { a => d.complete(a).void }
```

Attempts catch errors, not cancellations. When a task is cancelled, the fiber running the queue loop is cancelled with it, resulting in two things breaking simultaneously:

- d is never completed, so the caller of that task never returns. 
- The loop dies mid-iteration, so the state stays 'Active' and nothing else ever runs. This has been reproduced by three ignored specs, with one per file. Each one enqueues IO.cancelled, then a second task, and then awaits the second one. All three fail on the five-second suite timeout.

### Three shortcuts that do not work

Cats-Effect provides no means of detecting a cancellation as it occurs. Each of these was tried against the serial specification.

#### Uncancellable around the task 
IO.cancelled inside a mask is deferred, not swallowed. It lands as soon as the mask is lifted, before the attempt is made, so the fiber still dies. This also changes the meaning of the task: a task of IO.cancelled > ref.set(true) sets the ref.

queue advanced:              false
code after IO.canceled ran:  true

#### Attempt after uncancelable
Never sees the cancellation, for the reason above.

#### guaranteeCase alone.
This one is half a fix. Its finalizer runs on Outcome.Canceled(), so d is completed and the canceling task's own caller learns what happened instead of hanging. But `guaranteeCase` observes an outcome, it does not swallow it, so the cancellation still propagates and the queue is still wedged. Every later caller still hangs.

Stopping the propagation needs a boundary the cancellation cannot cross. That is either another fiber, or not having chained the other tasks behind it.

### Path A: a fiber per task

```scala
d <- Deferred[F, Outcome[F, Throwable, A]]
task = task0.start.flatMap { _.join }.flatMap { a => d.complete(a).void }
```

The runner observes the task's outcome instead of dying with it, and the caller's inner effect replays it with outcome.embed(F.canceled *> F.never).

Correct, and the same edit works in all three files. Costs:

benchmark | ratio
-- | --
Serial.pipelined, 256 tasks | 0.36x
Serial.pipelined, 16 tasks | 0.55x
Serial.sequential, 256 tasks | 0.70x
SerParQueue.keyless | 0.80x
[...]

PR #409, opened 2026-08-24
 Add a JMH module and a ConcurrentHashMap backed SerialKey

BrainHorse:
## Why the benchmark module?

Arguments about the serial queues were based on reading the code. Both fixes considered for https://github.com/evolution-gaming/cats-helper/issues/404 were rejected due to cost, neither cost was known until it was measured. As the module is not aggregated, neither `sbt test` nor CI ever run it, and the code as it stands is held in one baseline in `benchmark/README.md`.

## Why a second SerialKey?

The `SerialKey.of` function hash-partitions the keys into one `Ref[Map[K, Option[Task]]]` per core. Every enqueue and every advance rebuilds one of those maps. This is efficient when a few keys are active and the maps remain small. However, it becomes expensive when keys are spread, because the work per update increases with the number of live keys in a partition.

`ofConcurrentHashMap` maintains the same guarantees and queue loop while storing per-key state in a `ConcurrentHashMap` via `cats.effect.std.MapRef,` ensuring that an update only affects one entry.

## Summary by CodeRabbit

- **New Features**
  - Added an alternative per-key serialization implementation that allows tasks for different keys to run concurrently while preserving ordering for each key.
  - Expanded coverage to validate consistent behavior across serialization implementations.

- **Documentation**
  - Added guidance for running performance benchmarks, interpreting results, and updating baseline measurements.

- **Tests**
  - Added benchmark suites covering serial execution, keyed and keyless parallel queues, contention levels, and implementation comparisons.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -30,7 +30,8 @@ reading anything into a difference:
 | Command | `sbt "benchmark/Jmh/run -f 1 -wi 5 -i 5 -w 2s -r 2s .*Benchmark.*"` |
 
 Throughput in ops/s, higher is better. One `SerialKeyBenchmark` operation covers 64 tasks, so its
-scores are not comparable with the other two suites.
+scores are not comparable with the other two suites, and its rows are for
+`implementation=partitioned`.
 
 | Suite | Benchmark | keys | tasks | Score |
 | --- | --- | ---: | ---: | ---: |
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,56 @@
+# Benchmarks
+
+JMH benchmarks for `cats-helper`. The module is not part of the aggregate build, so `sbt test` and
+CI never run it. Run them by hand:
+
+```sh
+sbt benchmark/Jmh/run                                                       # everything
+sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerialKeyBenchmark"   # one suite
+sbt "benchmark/Jmh/run -prof gc SerialBenchmark.pipelined"                  # allocation too
+```
+
+`-f` forks, `-wi`/`-w` warmup iterations and duration, `-i`/`-r` measurement iterations and
+duration, `-p name=value` pins a `@Param`, `-rf json -rff out.json` saves results. These override
+the annotations on the class.
+
+Each suite says in its own scaladoc what it measures and why.
+
+## Baseline
+
+One measurement of the code as it stands, so a later change has something to be compared against.
+JMH scores are only comparable within one machine and one JDK, so reproduce this row before
+reading anything into a difference:
+
+| | |
+| --- | --- |
+| Commit | `c9eeb01` |
+| Machine | Apple M1 Pro, macOS 26.5.2 |
+| JDK | OpenJDK 25.0.3 |
+| Scala | 2.13.18 |
+| Command | `sbt "benchmark/Jmh/run -f 1 -wi 5 -i 5 -w 2s -r 2s .*Benchmark.*"` |
+
+Throughput in ops/s, higher is better.
+
+| Suite | Benchmark | keys | tasks | Score |
+| --- | --- | ---: | ---: | ---: |
+| `SerialBenchmark` | `pipelined` | | 1 | 100 335 ± 1 513 |
+| `SerialBenchmark` | `pipelined` | | 16 | 51 002 ± 1 941 |
+| `SerialBenchmark` | `pipelined` | | 256 | 9 297 ± 948 |
+| `SerialBenchmark` | `sequential` | | 1 | 90 810 ± 12 626 |
+| `SerialBenchmark` | `sequential` | | 16 | 37 408 ± 5 721 |
+| `SerialBenchmark` | `sequential` | | 256 | 3 151 ± 56 |
+| `SerialKeyBenchmark` | `singleThread` | 1 | | 100 411 ± 5 469 |
+| `SerialKeyBenchmark` | `singleThread` | 8 | | 97 471 ± 13 026 |
+| `SerialKeyBenchmark` | `singleThread` | 64 | | 95 699 ± 13 213 |
+| `SerialKeyBenchmark` | `eightThreads` | 1 | | 261 417 ± 12 493 |
+| `SerialKeyBenchmark` | `eightThreads` | 8 | | 277 859 ± 11 734 |
+| `SerialKeyBenchmark` | `eightThreads` | 64 | | 273 137 ± 24 661 |
+| `SerParQueueBenchmark` | `keyed` | 1 | | 261 167 ± 20 741 |
+| `SerParQueueBenchmark` | `keyed` | 8 | | 272 437 ± 26 488 |
+| `SerParQueueBenchmark` | `keyed` | 64 | | 274 495 ± 23 539 |
+| `SerParQueueBenchmark` | `keyless` | 1 | | 260 309 ± 17 944 |
+| `SerParQueueBenchmark` | `keyless` | 8 | | 255 453 ± 23 083 |
+| `SerParQueueBenchmark` | `keyless` | 64 | | 258 685 ± 11 854 |
+
+Replace this table when the baseline moves. The before and after of a single change belong in the
+description of the pull request that makes it.
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -28,59 +28,65 @@ reading anything into a difference:
 | Scala | 2.13.18 |
 | Command | `sbt "benchmark/Jmh/run -f 1 -wi 5 -i 5 -w 2s -r 2s .*Benchmark.*"` |
 
-Throughput in ops/s, higher is better. One `SerialKeyBenchmark` operation covers 64 tasks, so its
-scores are not comparable with the other two suites.
+Throughput in ops/s, higher is better. One operation covers 64 tasks in `SerialKeyBenchmark`,
+`SerParQueueBenchmark` and `SerParQueueKeylessBenchmark`, and `tasks` tasks in `SerialBenchmark`,
+so scores are only comparable within a suite.
 
 ### SerialBenchmark
 
 | Benchmark | tasks | Score |
 | --- | ---: | ---: |
-| `pipelined` | 1 | 103 279 ± 2 096 |
-| `pipelined` | 16 | 56 270 ± 514 |
-| `pipelined` | 256 | 9 483 ± 213 |
-| `sequential` | 1 | 107 029 ± 3 921 |
-| `sequential` | 16 | 39 737 ± 751 |
-| `sequential` | 256 | 3 164 ± 51 |
+| `pipelined` | 1 | 86 990 ± 12 361 |
+| `pipelined` | 16 | 52 630 ± 10 047 |
+| `pipelined` | 256 | 8 765 ± 163 |
+| `sequential` | 1 | 93 134 ± 24 153 |
+| `sequential` | 16 | 38 073 ± 5 130 |
+| `sequential` | 256 | 3 100 ± 69 |
+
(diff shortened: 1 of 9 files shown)


benchmark/src/main/scala/com/evolutiongaming/catshelper/SerParQueueBenchmark.scala
```diff
@@ -0,0 +1,55 @@
...
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Fork(1)
+@Warmup(iterations = 5, time = 2, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 2, timeUnit = TimeUnit.SECONDS)
+class SerParQueueBenchmark {
+
+  @Param(Array("1", "8", "64", "256", "1024"))
+  var keys: Int = 0
```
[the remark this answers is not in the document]
BrainHorse: Fixed in https://github.com/evolution-gaming/cats-helper/pull/409/changes/165974b4cacb0cdca7e0bc98839c4f5bbfc651e7

benchmark/src/main/scala/com/evolutiongaming/catshelper/SerialBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Cost of putting a task through [[Serial]] and waiting for its result.
+ *
+ * `pipelined` registers every task before awaiting any of them, so the queue runs a chained batch.
+ * `sequential` awaits each task before registering the next, so the queue is entered from idle
+ * every time. The two exercise the queue loop and the per-task wrapper in different proportions.
+ *
+ * To run: {{{sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerialBenchmark"}}}
+ */
```
mr-git, line 19: it is nice gesture to provide the anecdotal example with benchmark results, stating the machine and used JDK - specially, if any decisions are made based on results!

benchmark/src/main/scala/com/evolutiongaming/catshelper/SerialKeyBenchmark.scala
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * `keys` is the contention knob. `keys = 1` puts every thread on one key, larger counts spread the
+ * load over more of them. `implementation` picks between the two factories `SerialKey` offers.
+ *
+ * One operation enqueues `tasksPerOp` tasks and awaits them all. Enqueueing a single task per
+ * operation measures the handoff between the calling thread and the compute pool rather than the
+ * queue, which on this workload swamps the result with noise.
+ *
+ * To run: {{{sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerialKeyBenchmark"}}}
+ */
```
mr-git, line 23: same as for `SerialBenchmark` - some results are welcome

benchmark/src/main/scala/com/evolutiongaming/catshelper/SerParQueueBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * The keyless case has no key to vary, so it lives in [[SerParQueueKeylessBenchmark]] rather than
+ * repeating once per `keys` value here.
+ *
+ * One operation enqueues `tasksPerOp` tasks and awaits them all. Enqueueing a single task per
+ * operation measures the handoff between the calling thread and the compute pool rather than the
+ * queue, which on this workload swamps the result.
+ *
+ * To run: {{{sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerParQueueBenchmark"}}}
+ */
```
mr-git, line 23: please add example results

benchmark/src/main/scala/com/evolutiongaming/catshelper/SerParQueueKeylessBenchmark.scala
```diff
@@ -0,0 +1,51 @@
...
+ * Cost of putting a keyless task through [[SerParQueue]], which orders against every other task.
+ *
+ * Separate from [[SerParQueueBenchmark]] because a keyless task has no key, so a `keys` parameter
+ * would only repeat the same measurement.
+ *
+ * One operation enqueues `tasksPerOp` tasks and awaits them all, for the reason given in
+ * [[SerParQueueBenchmark]].
+ *
+ * To run: {{{sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerParQueueKeylessBenchmark"}}}
+ */
```
mr-git, line 21: results

benchmark/README.md
```diff
@@ -0,0 +1,92 @@
...
+sbt benchmark/Jmh/run                                                       # everything
+sbt "benchmark/Jmh/run com.evolutiongaming.catshelper.SerialKeyBenchmark"   # one suite
+sbt "benchmark/Jmh/run -prof gc SerialBenchmark.pipelined"                  # allocation too
+```
+
+`-f` forks, `-wi`/`-w` warmup iterations and duration, `-i`/`-r` measurement iterations and
+duration, `-p name=value` pins a `@Param`, `-rf json -rff out.json` saves results. These override
+the annotations on the class.
+
+Each suite says in its own scaladoc what it measures and why.
```
mr-git:
```suggestion
Each suite says in its own Scaladoc what it measures and why.
```

benchmark/README.md
```diff
@@ -0,0 +1,92 @@
...
+One measurement of the code as it stands, so a later change has something to be compared against.
+JMH scores are only comparable within one machine and one JDK, so reproduce this row before
+reading anything into a difference:
+
+| | |
+| --- | --- |
+| Machine | Apple M1 Pro, macOS 26.5.2 |
+| JDK | OpenJDK 25.0.3 |
+| Scala | 2.13.18 |
+| Command | `sbt "benchmark/Jmh/run -f 1 -wi 5 -i 5 -w 2s -r 2s .*Benchmark.*"` |
```
mr-git, line 29: it would be nicer, if table would be readable in text, without markdown compilation (IDEA provides means to reformat table easily)

benchmark/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Throughput in ops/s, higher is better. One operation covers 64 tasks in `SerialKeyBenchmark`,
+`SerParQueueBenchmark` and `SerParQueueKeylessBenchmark`, and `tasks` tasks in `SerialBenchmark`,
+so scores are only comparable within a suite.
+
+### SerialBenchmark
+
+| Benchmark | tasks | Score |
```
mr-git, line 37: please format *ALL* tables for improved readability in their raw format

benchmark/README.md
```diff
@@ -0,0 +1,92 @@
...
+| `pipelined` | 256 | 8 765 ± 163 |
+| `sequential` | 1 | 93 134 ± 24 153 |
+| `sequential` | 16 | 38 073 ± 5 130 |
+| `sequential` | 256 | 3 100 ± 69 |
+
+The `tasks = 1` rows carry a wide error because one operation is a single task, so the measurement
+is dominated by the handoff between the calling thread and the compute pool. Read the 256 rows for
+a stable figure.
+
+### SerialKeyBenchmark
```
mr-git, line 50: for all benchmarks - please add conclusions, the Next Developer might not understand what you saw here and might drive conflicting conclusions

core/src/main/scala/com/evolutiongaming/catshelper/SerialKey.scala
```diff
@@ -117,4 +118,84 @@ object SerialKey {
...
+        def start(key: K, task: Task): F[Unit] = {
+          task
+            .tailRecM { task =>
+              task *> mapRef(key).modify {
+                case Some(KeyState.Pending(tasksChain)) =>
+                  (KeyState.Running.some, tasksChain.asLeft[Unit])
+                case Some(KeyState.Running) =>
+                  (none, ().asRight[Task])
+                case None =>
+                  (none, ().asRight[Task])
```
mr-git: just curious, why not like following?:
```suggestion
                case _ =>
                  (none, ().asRight[Task])
```

core/src/test/scala/com/evolutiongaming/catshelper/SerialKeyTest.scala
```diff
@@ -20,123 +20,131 @@ class SerialKeyTest extends AsyncFunSuite with Matchers {
...
+  implementations[String].foreach { case (name, serialKey) =>
+    test(s"$name runs tasks of one key in submission order") {
+      val tasks = 200
+      val result = for {
+        serial <- serialKey
+        order <- Ref[IO].of(List.empty[Int])
+        awaits <- (1 to tasks).toList.traverse { index => serial("key") { order.update { index :: _ } } }
+        _ <- awaits.sequence_
+        observed <- order.get
+        _ <- IO { observed.reverse shouldEqual (1 to tasks).toList }
```
mr-git: why have to `.reverse`?

mr-git (approved):
LGTM

benchmark/src/main/scala/com/evolutiongaming/catshelper/SerialBenchmark.scala
BrainHorse: The results are in the `benchmark/README.md` file, together with information on the machine, JDK and Scala versions, as well as the exact command. It is this information that makes the results meaningful.
Copying figures into four Scaladoc files means four copies to update, and in practice, these would not be synced. Scaladoc also has the wrong lifespan for a number that is valid for one machine and one JDK.

benchmark/README.md
BrainHorse: Since we last discussed this matter (https://github.com/evolution-gaming/play-json-tools/pull/44#discussion_r3710338088), my opinion hasn't changed. Padding columns would need to be redone with every update and will affect all columns, not only the real change. GitHub as well as other modern viewers/editors render the tables and align them when perfectly. I would prefer the file to remain inexpensive to regenerate.

> (IDEA provides means to reformat table easily)

I would prefer not to install the IDEA for a single purpose of prettifying MD tables.

benchmark/README.md
BrainHorse: I believe the only conclusion that is relevant here is already written:

> Throughput in ops/s, higher is better.

Decisions that drive conclusions are better suited to places where they are relevant and have a fixed lifetime, such as in the PR description of a change and in the Scaladoc of the selected option. `SerialKey.ofConcurrentHashMap` explains the circumstances in which it should be preferred to `of`, and the reference to the beanchmarks

core/src/test/scala/com/evolutiongaming/catshelper/SerialKeyTest.scala
BrainHorse: The `.reverse` was there because `index :: _` prepends. Prepending 200 tasks leaves the list in newest-first order, so reverse puts it back in submission order for comparison. From the spec perspective, it was the correct behaviour.  
Removed in https://github.com/evolution-gaming/cats-helper/pull/409/changes/41a4ccab9a566821db67faf81c7b70da2a034e11 in favour of append

benchmark/README.md
mr-git: we are not writing text just for bots to read - humans read them and in many cases the tables are read in their raw textual form, like in git diffs, not over some renderer.

BrainHorse: image
We do have renderers for human readable form in many modern systems, including git diff. 
Unless we have md tables prettifier/checker configured at a repository level, which would work independently of the IDE, like we do have it with scalafmt, I wouldn't agree that adding spaces to tables is worthwhile.

PR #414, opened 2026-08-26
Fix flaky `CountLatchSpec`

### LINKED ISSUE #399: flaky unit-test `CountLatchSpec`
noticed in: https://github.com/evolution-gaming/cats-helper/actions/runs/31432347714/job/93598505322?pr=391
```
[info] CountLatchSpec:
[info] - init with 0 *** FAILED ***
[info]   false was not equal to true (CountLatchSpec.scala:22)
```

BrainHorse:
Fix https://github.com/evolution-gaming/cats-helper/issues/399
Root cause: The CountLatchSpec's blocked/done helpers used `timeoutTo(1 millisecond, ...)` on a real wall clock, competing with `IORuntime.global` (shared by every suite in the module) against a 1 ms timer. Under CI load/scheduler contention, this race could occasionally be lost, even for a trivially resolved latch, resulting in the reported false positive.

Fix: The tests are run under `cats.effect.testkit.TestControl.executeEmbed` instead of the real `IORuntime`. This is a single-threaded, deterministic scheduler with virtual time, so the 1 ms timeout only 'elapses' when nothing else is runnable. This correctly separates `resolves instantly` from `blocks forever` with zero dependency on real CPU scheduling.

## Summary by CodeRabbit

* **Tests**
  * Improved CountLatch test execution using controlled effect simulation for more reliable and deterministic test results.

### DIFF
--- a/core/src/test/scala/com/evolutiongaming/catshelper/CountLatchSpec.scala
+++ b/core/src/test/scala/com/evolutiongaming/catshelper/CountLatchSpec.scala
@@ -114,7 +114,7 @@ class CountLatchSpec extends AnyFunSuite with Matchers {
       blocked1 shouldBe true
       done shouldBe true
     }
-    io.unsafeRunSync()
+    TestControl.executeEmbed(io).unsafeRunSync()
   }
 
   test("concurrent acquire & release") {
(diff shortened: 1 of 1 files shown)


mr-git:
LGTM, please adjust wording in description, the `Fix: The tests **were** run under...` is wrong

mr-git (approved):
LGTM

PR #415, opened 2026-08-27
Guard CountLatch.release against non-positive counts

BrainHorse:
Unlike acquire which ignores calls with `n < 1`, `release(n: Int)` had no lower bound check. Passing a negative to release could silently increases internal latch counter.

This change means that release now ignores `n < 1`, which is consistent with the existing behaviour for acquire. Both now use `F.whenA(n ≥ 1)` for the guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved latch behavior when acquiring or releasing with nonpositive counts.
  * Negative releases no longer incorrectly affect latch completion.

* **Tests**
  * Added coverage confirming that ignored negative releases leave the latch blocked until a valid release occurs.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/CountLatch.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/CountLatch.scala
@@ -77,8 +77,7 @@ object CountLatch {
       val F = Async[F]
 
       override def acquire(n: Int): F[Unit] =
-        if (n < 1) F.unit
-        else
+        F.whenA(n >= 1) {
           state.access
             .flatMap {
               case (state, set) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/evolutiongaming/catshelper/CountLatch.scala
```diff
@@ -93,17 +92,18 @@ object CountLatch {
...
       override def release(n: Int): F[Unit] =
-        state
-          .modify {
-            case Done => Done -> F.unit
-            case Awaiting(l, await) =>
-              if (l > n) Awaiting(l - n, await) -> F.unit
-              else Done -> await.complete(()).void
-          }
-          .flatten
-          .uncancelable
```
mr-git, line 106: why `.uncancelable` is removed?

BrainHorse: It is a part of `flatModify`. I have replaced the `.modify {...}.flatten.uncancelable` with `flatModify`

mr-git: ok, didn't know that. thanks!

mr-git (approved):

PR #421, opened 2026-08-31
Add ToTryTimeoutSpec: pin what #330 fixes

tobiajo:
Unit tests for the `ioToTry` timeout, on top of #330.

On master, `ioToTry` uses `syncStep` + `unsafeRunTimed`. On timeout `unsafeRunTimed` fire-and-forgets a cancel and returns immediately — it never waits for finalizers, and `.uncancelable` work keeps running. kafka-flow's [`TopicFlow.safeguard`](https://github.com/evolution-gaming/kafka-flow/blob/master/core/src/main/scala/com/evolutiongaming/kafka/flow/TopicFlow.scala#L181) takes a semaphore permit inside `.uncancelable`; the poll thread moves on while the abandoned fiber still holds the permit, and every subsequent `poll` blocks on the semaphore — the consumer silently stops processing.

All three specs fail on master, pass on this branch:

| spec | master | #330 |
|---|:-:|:-:|
| timeout releases a resource acquired before the async boundary | ✗ | ✓ |
| a timed-out attempt does not strand the guard it acquired | ✗ | ✓ |
| an uncancelable guarded recovery completes and releases its guard | ✗ | ✓ |

To see both directions:

\`\`\`
git checkout master -- core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
sbt "core/testOnly *ToTryTimeoutSpec"
\`\`\`

PR #330, opened 2025-03-28
use `IO`'s runtime directly for `ioToTry` and `ioToFuture`

mr-git:
alternative fix for https://github.com/evolution-gaming/skafka/pull/477:

<details>
  <summary>Summary from `skafka` PR</summary>
  Subscribing to thousands of partition-topics is not a problem, but when `RebalanceListener1` is added, consumer fails to init with `StackOverflowException`, like:

```
// thousands of next line with increasing index
interpret:2105, SyncStep$ (cats.effect) [2]
interpret:2105, SyncStep$ (cats.effect) [1]
syncStep:1939, IO$$anon$2 (cats.effect)
syncStep:1935, IO$$anon$2 (cats.effect)
syncStep:1047, IO (cats.effect)
apply$$anonfun$1:44, ToTry$$anon$4 (com.evolutiongaming.catshelper)
apply:-1, ToTry$$anon$4$$Lambda/0x0000007e01744318 (com.evolutiongaming.catshelper)
apply:217, Try$ (scala.util)
apply:50, ToTry$$anon$4 (com.evolutiongaming.catshelper)
apply:38, ToTry$$anon$4 (com.evolutiongaming.catshelper)
toTry$extension:74, CatsHelper$OpsCatsHelper$ (com.evolutiongaming.catshelper)
com$evolutiongaming$skafka$consumer$ConsumerConverters$RebalanceListener1Ops$$$_$onPartitions$2:141, ConsumerConverters$RebalanceListener1Ops$ (com.evolutiongaming.skafka.consumer)
onPartitionsAssigned:151, ConsumerConverters$RebalanceListener1Ops$$anon$2 (com.evolutiongaming.skafka.consumer)
invokePartitionsAssigned:65, ConsumerRebalanceListenerInvoker (org.apache.kafka.clients.consumer.internals)
onJoinComplete:425, ConsumerCoordinator (org.apache.kafka.clients.consumer.internals)
joinGroupIfNeeded:504, AbstractCoordinator (org.apache.kafka.clients.consumer.internals)
ensureActiveGroup:415, AbstractCoordinator (org.apache.kafka.clients.consumer.internals)
poll:511, ConsumerCoordinator (org.apache.kafka.clients.consumer.internals)
updateAssignmentMetadataIfNeeded:657, ClassicKafkaConsumer (org.apache.kafka.clients.consumer.internals)
poll:616, ClassicKafkaConsumer (org.apache.kafka.clients.consumer.internals)
[...]

PR #425, opened 2026-09-08
Make `ioToTry`'s timeout respect `uncancelable` and run finalizers

tobiajo:
Supersedes #330's work on `ioToTry` and incorporates #421's tests pinning down whether `Resource` finalizers (release) complete after a timeout.

The `StackOverflowException` behind evolution-gaming/skafka#477 was fixed upstream in cats-effect 3.6.2; it is pinned here by the 100k-deep chain test. Incidentally, #330 also fixes the `Resource` finalizer issue by removing `syncStep` entirely.

## Why

The main difference between master, #330 and this PR is that master and this PR use `syncStep`, while #330 does not.

Without `syncStep`, the whole effect runs through the Cats Effect runtime, adding the overhead of starting and synchronously waiting for a fiber even when the effect could complete synchronously.

Also, #330 makes synchronous effects subject to the timeout, while master and this PR avoid this.

## Unlawful `SyncIO`

The debatable complexity added by this PR is a custom `Sync` instance around `SyncIO`. It extends `Sync[SyncIO]`, only overriding `rootCancelScope` from `Uncancelable` to `Cancelable`.

Looking under the hood in cats-effect [here](https://github.com/typelevel/cats-effect/blob/v3.7.1/core/shared/src/main/scala/cats/effect/IO.scala#L2390-L2399) is where the `rootCancelScope` makes a difference:

```scala
case IO.Uncancelable(body, _) if G.rootCancelScope == CancelScope.Uncancelable =>
  // walk inside: unwrap the region, replace Poll with identity, keep going
  val ioa = body(new Poll[IO] { def apply[C](ioc: IO[C]): IO[C] = ioc })
  interpret(ioa, limit, stepsUntilDefer)

case IO.OnCancel(ioa, _) if G.rootCancelScope == CancelScope.Uncancelable =>
  // walk past: keep the effect, drop the finalizer
  interpret(ioa, limit, stepsUntilDefer)

case _ =>
  // stop: return the node as is
  G.pure(Left(io))
```

Because `Sync[SyncIO]` is `Uncancelable`, `syncStep` may step through `uncancelable`, `onCancel`, and `Resource` handling. If it later stops at an async boundary, the remaining `IO` may have lost its finalizers, so a timeout may skip the release entirely.

The custom instance reports itself as `Cancelable`, so `syncStep` falls through to `case _` at `Uncancelable` and `OnCancel` instead of entering them, and returns the remaining `IO` with its finalizers intact. It is unlawful because `SyncIO` cannot actually be cancelled, so it is kept private and used only to control where `syncStep` stops.

## Why `IO.timeout` instead of `unsafeRunTimed`

`ToTry` returns a `Try` the caller acts on immediately. `IO.timeout` guarantees that the effect is done (cancelled with finalizers run, or completed) before returning. `unsafeRunTimed` does not: it returns while the finalizer is still running on a detached fiber.

## Verdict

In short, we either accept the extra complexity of this PR or the added runtime overhead of #330, which also makes synchronous effects subject to the timeout.

## Bug Fixes

- Pure and delay-based operations now execute synchronously when possible.
- Asynchronous operations have more reliable timeout and cancellation handling.
- Resources, permits, and cancellation guards are correctly released after timeouts.
- Timeout processing waits for required cleanup before completing, enabling safe recovery and retries.

## Tests

- Expanded coverage for synchronous execution, asynchronous boundaries, timeouts, cancellation, resource cleanup, and recovery scenarios.

The leak this closes is described from the kafka-flow side in evolution-gaming/kafka-flow#937.

### DIFF
--- a/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
@@ -35,15 +35,14 @@ object ToTry {
    * Please think twice before using this, ideally you should not have toTry in your `pure` code
    * base!
    *
-   * Simple effects (`pure`, `delay`, `map`, `flatMap`, `attempt`, `handleErrorWith`, and therefore
-   * `Ref` operations and `Deferred#complete`) run on the calling thread and never time out. From
-   * the first `uncancelable`, `onCancel`, `Resource` or asynchronous boundary onwards the effect
-   * runs as a fiber under
+   * Simple effects (`pure`, `delay`, `map`, `flatMap`, `attempt`, `handleErrorWith`) run on the
+   * calling thread and never time out. From the first `uncancelable`, `onCancel`, `Resource` or
+   * asynchronous boundary onwards the effect runs as a fiber under
    * [[https://typelevel.org/cats-effect/docs/datatypes/io#scalatimeout IO.timeout]].
    *
    * @param timeout
-   *   applies to the part that runs as a fiber. On expiry the fiber is cancelled (not abandoned),
-   *   its finalizers run, and the result is `Failure(TimeoutException)`. Inside an
+   *   applies to the fiber part. On expiry the fiber is cancelled (not abandoned), its finalizers
+   *   run, and the result is `Failure(TimeoutException)`. Inside an
    *   [[https://typelevel.org/cats-effect/docs/typeclasses/monadcancel#uncancelable-regions uncancelable region]]
    *   there is nothing to cancel, so the effect runs to completion regardless of the timeout.
    */
@@ -75,23 +74,18 @@ object ToTry {
   }
 
   /**
-   * A `Sync[SyncIO]` whose only purpose is to control where
+   * Controls where
    * [[https://typelevel.org/cats-effect/api/3.x/cats/effect/kernel/Async.html#syncStep syncStep]]
-   * stops walking.
+   * stops walking an `IO`.
    *
-   * `syncStep` walks an `IO` node by node on the calling thread and returns whatever it could not
-   * walk as a new `IO`. How far it walks depends on `rootCancelScope` of the `Sync` instance it is
-   * given. With `SyncIO`'s own instance (scope = `Uncancelable`) it walks inside `uncancelable`
-   * regions and past `onCancel` finalizers, because `SyncIO` can never be cancelled anyway. The
-   * `IO` it returns then lacks those protections, and a timeout cancelling it skips the finalizers.
+   * `syncStep` walks an `IO` node by node on the calling thread. How far it walks depends on
+   * `rootCancelScope` of the `Sync` it is given. With `SyncIO`'s own instance (scope =
+   * `Uncancelable`) it walks inside `uncancelable` and past `onCancel`, so the returned `IO` lacks
+   * those protections. This instance reports scope = `Cancelable`, so `syncStep` stops there and
+   * returns the region whole.
    *
-   * This instance reports scope = `Cancelable`, so `syncStep` stops at `uncancelable` and
-   * `onCancel` and returns the region as is, protections included.
-   *
-   * `rootCancelScope` is the only member `syncStep` consults for this decision. Every other member
-   * delegates to `SyncIO`'s own `Sync`. The instance is unlawful (`SyncIO` cannot actually be
-   * cancelled, which the `Cancelable` scope falsely promises) and therefore private and never
-   * implicit.
+   * Only `rootCancelScope` matters; every other member delegates to `SyncIO`'s own `Sync`. The
+   * instance is unlawful (`SyncIO` cannot actually be cancelled), hence private and never implicit.
    *
    * @see
    *   [[https://typelevel.org/cats-effect/docs/typeclasses/monadcancel MonadCancel]] for
--- a/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
@@ -2,12 +2,12 @@ package com.evolutiongaming.catshelper
 
 import cats.Id
 import cats.arrow.FunctionK
-import cats.effect.IO
+import cats.effect.kernel.{CancelScope, Poll, Sync}
 import cats.effect.unsafe.IORuntime
+import cats.effect.{IO, SyncIO}
 
-import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
-import scala.util.{Failure, Success, Try}
+import scala.util.{Success, Try}
 
 trait ToTry[F[_]] {
 
@@ -35,9 +35,17 @@ object ToTry {
    * Please think twice before using this, ideally you should not have toTry in your `pure` code
    * base!
    *
+   * Runs the effect on the calling thread for as long as it is pure or `delay`-shaped (`pure`,
+   * `delay`, `map`, `flatMap`, `attempt`, `handleErrorWith`, hence also `Ref` operations and
+   * `Deferred#complete`). Such effects complete inline, never touch the runtime and never time out.
+   * At the first asynchronous boundary, `uncancelable`, `onCancel` or `Resource` allocation, the
+   * rest of the effect runs as a fiber on the runtime under `IO.timeout`.
+   *
    * @param timeout
-   *   used only for computation after first seen async boundary, covering all computations onwards
-   *   in case there is no async boundary found, timeout is not used
+   *   bound on the part that runs on the runtime. It is cooperative: where the effect is cancelable
+   *   the fiber is cancelled, its finalizers run and the result is `Failure(TimeoutException)`;
+   *   inside an `uncancelable` region it waits for the region to end, so a masked effect can run
+   *   past it.
    */
   def ioToTry(
     timeout: FiniteDuration,
@@ -45,21 +53,11 @@ object ToTry {
     runtime: IORuntime,
   ): ToTry[IO] = new ToTry[IO] {
 
-    def apply[A](fa: IO[A]) = {
-
-      def error: Try[A] = Failure[A](new TimeoutException(timeout.toString()))
-
-      for {
-        a <- Try {
-          fa.syncStep(Int.MaxValue).unsafeRunSync() match {
-            case Left(computation) =>
-              computation.unsafeRunTimed(timeout)
-            case Right(value) =>
-              Some(value)
-          }
-        }
-        a <- a.fold(error) { a => Success(a) }
-      } yield a
+    def apply[A](fa: IO[A]) = Try {
+      IO.asyncForIO.syncStep[SyncIO, A](fa, Int.MaxValue)(CancelableSyncIO).unsafeRunSync() match {
+        case Right(a) => a
+        case Left(rest) => rest.timeout(timeout).unsafeRunSync()
+      }
     }
   }
 
@@ -74,5 +72,47 @@ object ToTry {
 
   implicit val tryToTry: ToTry[Try] = new ToTry[Try] {
     def apply[A](fa: Try[A]) = fa
+  }
+
+  /**
+   * `Sync[SyncIO]` that reports a `Cancelable` root scope.
+   *
+   * `syncStep` steps into `uncancelable` and `onCancel` only when the target's root scope is
+   * `Uncancelable`, handing back a remainder without its mask and finalizers. Under this instance
+   * it stops at those nodes with the remainder intact.
+   *
+   * Unlawful for `SyncIO`, hence private and never implicit. The stepper only uses `pure`,
+   * `raiseError`, `delay`, `defer`, `realTime`, `monotonic`, `map`, `flatMap`, `handleError` and
+   * `handleErrorWith`, so the wrong scope is never observed.
+   */
+  private object CancelableSyncIO extends Sync[SyncIO] {
+
+    private val F = SyncIO.syncForSyncIO
+
+    def rootCancelScope: CancelScope = CancelScope.Cancelable
+
+    def pure[A](a: A): SyncIO[A] = F.pure(a)
+
+    def raiseError[A](e: Throwable): SyncIO[A] = F.raiseError(e)
+
+    def handleErrorWith[A](fa: SyncIO[A])(f: Throwable => SyncIO[A]): SyncIO[A] = F.handleErrorWith(fa)(f)
+
+    def flatMap[A, B](fa: SyncIO[A])(f: A => SyncIO[B]): SyncIO[B] = F.flatMap(fa)(f)
+
+    def tailRecM[A, B](a: A)(f: A => SyncIO[Either[A, B]]): SyncIO[B] = F.tailRecM(a)(f)
+
+    def suspend[A](hint: Sync.Type)(thunk: => A): SyncIO[A] = F.suspend(hint)(thunk)
+
+    def monotonic: SyncIO[FiniteDuration] = F.monotonic
+
+    def realTime: SyncIO[FiniteDuration] = F.realTime
+
+    def forceR[A, B](fa: SyncIO[A])(fb: SyncIO[B]): SyncIO[B] = F.forceR(fa)(fb)
+
+    def uncancelable[A](body: Poll[SyncIO] => SyncIO[A]): SyncIO[A] = F.uncancelable(body)
+
+    def canceled: SyncIO[Unit] = F.canceled
+
+    def onCancel[A](fa: SyncIO[A], fin: SyncIO[Unit]): SyncIO[A] = F.onCancel(fa, fin)
   }
 }
--- a/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
+++ b/core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
@@ -35,17 +35,17 @@ object ToTry {
    * Please think twice before using this, ideally you should not have toTry in your `pure` code
    * base!
    *
-   * Runs the effect on the calling thread for as long as it is pure or `delay`-shaped (`pure`,
-   * `delay`, `map`, `flatMap`, `attempt`, `handleErrorWith`, hence also `Ref` operations and
-   * `Deferred#complete`). Such effects complete inline, never touch the runtime and never time out.
-   * At the first asynchronous boundary, `uncancelable`, `onCancel` or `Resource` allocation, the
-   * rest of the effect runs as a fiber on the runtime under `IO.timeout`.
+   * A pure or `delay`-shaped effect (`pure`, `delay`, `map`, `flatMap`, `attempt`,
+   * `handleErrorWith`, hence also `Ref` operations and `Deferred#complete`) runs on the calling
+   * thread: it never reaches the runtime and never times out. From the first asynchronous boundary,
+   * `uncancelable`, `onCancel` or `Resource` allocation onwards, the effect runs as a fiber under
+   * `IO.timeout`.
    *
    * @param timeout
-   *   bound on the part that runs on the runtime. It is cooperative: where the effect is cancelable
-   *   the fiber is cancelled, its finalizers run and the result is `Failure(TimeoutException)`;
-   *   inside an `uncancelable` region it waits for the region to end, so a masked effect can run
-   *   past it.
+   *   bound on the part that runs as a fiber. `IO.timeout` cancels rather than abandons: where the
+   *   effect is cancelable it stops, its finalizers run, and the result is
+   *   `Failure(TimeoutException)`. Cancellation waits for an `uncancelable` region to end, so an
+   *   effect can run past the timeout inside one.
    */
   def ioToTry(
     timeout: FiniteDuration,
@@ -53,7 +53,7 @@ object ToTry {
     runtime: IORuntime,
   ): ToTry[IO] = new ToTry[IO] {
 
-    def apply[A](fa: IO[A]) = Try {
+    def apply[A](fa: IO[A]): Try[A] = Try {
       IO.asyncForIO.syncStep[SyncIO, A](fa, Int.MaxValue)(CancelableSyncIO).unsafeRunSync() match {
         case Right(a) => a
         case Left(rest) => rest.timeout(timeout).unsafeRunSync()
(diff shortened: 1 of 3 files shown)


core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  test("pure and delay-shaped effects never touch the runtime") {
+    val effect = IO.pure(1).map(_ + 1).flatMap(x => IO(x * 2)).handleErrorWith(_ => IO.pure(-1))
+
+    toTryNoRuntime(effect) shouldEqual Success(4)
+  }
+
+  test("a 100k-deep flatMap chain over Ref.update steps to completion without a fiber") {
+    val effect = IO.ref(0).flatMap { ref =>
+      (1 to 100000).foldLeft(IO.unit)((acc, _) => acc.flatMap(_ => ref.update(_ + 1))) *> ref.modify(n => (n, n))
```
tobiajo:
```suggestion
      (1 to 100000).foldLeft(IO.unit)((acc, _) => acc.flatMap(_ => ref.update(_ + 1))) *> ref.get
```

Current was a workaround to get Codacy pass.

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -75,4 +74,60 @@ object ToTry {
   implicit val tryToTry: ToTry[Try] = new ToTry[Try] {
     def apply[A](fa: Try[A]) = fa
   }
+
+  /**
+   * `SyncIO` under a `Sync` instance whose `rootCancelScope` is `Cancelable`.
+   *
+   * `syncStep` steps into `uncancelable` and `onCancel` only when the target's root scope is
+   * `Uncancelable`, handing back a remainder without its mask and finalizers. Under this instance
+   * it stops at those nodes with the remainder intact.
```
mr-git: I do not get this! What `mask`, whose `mask`? Is `this` the same as `its` and `it` while being different from `those`?

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -75,4 +74,60 @@ object ToTry {
...
+   */
+  private final class CancelableSyncIO[A](val value: SyncIO[A])
+
+  private object CancelableSyncIO extends Sync[CancelableSyncIO] {
+
+    private val F = SyncIO.syncForSyncIO
+
+    private def wrap[A](fa: SyncIO[A]): CancelableSyncIO[A] = new CancelableSyncIO(fa)
+
+    def rootCancelScope: CancelScope = CancelScope.Cancelable
```
mr-git: nitpicking: I'd use the same "shape" for all `def`s, like:
```scala
def xxx: Type =
  // xxx's block
```

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -75,4 +74,60 @@ object ToTry {
...
   }
+
+  /**
+   * `SyncIO` under a `Sync` instance whose `rootCancelScope` is `Cancelable`.
+   *
+   * `syncStep` steps into `uncancelable` and `onCancel` only when the target's root scope is
+   * `Uncancelable`, handing back a remainder without its mask and finalizers. Under this instance
+   * it stops at those nodes with the remainder intact.
+   *
+   * Unlawful for `SyncIO`, hence private. The stepper only uses `pure`, `raiseError`, `delay`,
```
mr-git: Why is it `unlawful`?

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -75,4 +74,60 @@ object ToTry {
...
+
+    def forceR[A, B](fa: CancelableSyncIO[A])(fb: CancelableSyncIO[B]): CancelableSyncIO[B] =
+      wrap(F.forceR(fa.value)(fb.value))
+
+    def uncancelable[A](body: Poll[CancelableSyncIO] => CancelableSyncIO[A]): CancelableSyncIO[A] =
+      wrap(F.uncancelable { poll =>
+        body(new Poll[CancelableSyncIO] {
+          def apply[B](fb: CancelableSyncIO[B]): CancelableSyncIO[B] = wrap(poll(fb.value))
+        }).value
+      })
```
mr-git: If this is the "salt" of the `CancelableSyncIO`, how is it better than whatever we had before? Where is the `remainder without its mask and finalizers`?

mr-git (commented):
I found about those `cats effect masks`!!! :
* https://github.com/typelevel/cats-effect/blob/bb3748611529c7d4225d129d26fb499862c20213/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala#L70-L91
* https://typelevel.org/cats-effect/api/3.x/cats/effect/kernel/Async.html#uncancelable[A](body:cats.effect.kernel.Poll[F]=%3EF[A]):F[A]
* https://github.com/typelevel/cats-effect/blob/bb3748611529c7d4225d129d26fb499862c20213/docs/concepts.md?plain=1#L104

It would be nicer if we would use the human-language at our custom API, instead of relying on readers knowledge of Cats-Effect specific internal nuances. We should at least reference to them, in Scaladoc explicitly provide link(s) to the concepts we use and provide details on `why?` we implement something custom.

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    toTry(effect) should matchPattern { case Failure(_: TimeoutException) => }
+
+    acquired.get() shouldBe true
+    released.get() shouldBe true
+  }
+
+  private object RuntimeTouched extends RuntimeException("IORuntime was touched") with NoStackTrace
+
+  private def poisoned: IORuntime = {
```
mr-git: Can we have names, which are not AI-poisoned (watermarked)?

Sarcastic (and funny!!!) suggestion (sorry):
```suggestion
  private def loadUnBearing: IORuntime = {
```

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
mr-git: for easier parsing
```suggestion
      (1 to 100_000).foldLeft(IO.unit)((acc, _) => acc.flatMap(_ => ref.update(_ + 1))) *> ref.modify(n => (n, n))
```

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    toTryNoRuntime(effect) shouldEqual Success(4)
+  }
+
+  test("a 100k-deep flatMap chain over Ref.update steps to completion without a fiber") {
+    val effect = IO.ref(0).flatMap { ref =>
+      (1 to 100000).foldLeft(IO.unit)((acc, _) => acc.flatMap(_ => ref.update(_ + 1))) *> ref.modify(n => (n, n))
+    }
+
+    toTryNoRuntime(effect) shouldEqual Success(100000)
```
mr-git:
```suggestion
    toTryNoRuntime(effect) shouldEqual Success(100_000)
```

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  for {
+    (name, effect, inline) <- List(
+      ("Ref#update", IO.ref(0).flatMap(_.update(_ + 1)), true),
+      ("Deferred#complete", IO.deferred[Int].flatMap(_.complete(1).void), true),
+      ("uncancelable", IO.unit.uncancelable, false),
+      ("onCancel", IO.unit.onCancel(IO.unit), false),
+      ("Resource.allocated", Resource.make(IO.unit)(_ => IO.unit).allocated.void, false),
```
mr-git, line 47: About all these `false` cases - we test that they "touch the IORuntime", but I cannot understand why it is expected, or what does it prove.

What about case where we would have structure something like `.uncancelable(.cancelable(.uncancelable(.cancelable))))` - what is an expected behaviour? In `skafka`'s `RebalanceListener` we could have a structure, which could be similar to before-mentioned one, with `Semaphore` added to a mix. Do we want or attempt to unit-test that `skafka`'s structure with `kafka-flow`'s wrapping(s) here too?

tobiajo:
> I found about those `cats effect masks`!!! :
> 
> * https://github.com/typelevel/cats-effect/blob/bb3748611529c7d4225d129d26fb499862c20213/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala#L70-L91
> * [https://typelevel.org/cats-effect/api/3.x/cats/effect/kernel/Async.html#uncancelable[A](body:cats.effect.kernel.Poll[F]=%3EF[A]):F[A]](https://typelevel.org/cats-effect/api/3.x/cats/effect/kernel/Async.html#uncancelable%5BA%5D(body:cats.effect.kernel.Poll%5BF%5D=%3EF%5BA%5D):F%5BA%5D)
> * https://github.com/typelevel/cats-effect/blob/bb3748611529c7d4225d129d26fb499862c20213/docs/concepts.md?plain=1#L104
> 
> It would be nicer if we would use the human-language at our custom API, instead of relying on readers knowledge of Cats-Effect specific internal nuances. We should at least reference to them, in Scaladoc explicitly provide link(s) to the concepts we use and provide details on `why?` we implement something custom.

It is debatable if the `CancelableSyncIO` is worth it, I'll try to showcase side by side how differs from #330. There might also be some middle-ground. Either is needed though, current `ioToTry` is a trap when finalizers not called.

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
mr-git: never mind, found it later in next spec

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+ * without waiting: the permit was never released and every subsequent poll blocked on the
+ * semaphore. The consumer silently stopped processing.
+ *
+ * All three specs fail against the previous `ioToTry` and pass with the cooperative timeout.
+ */
+class ToTryTimeoutSpec extends AnyFunSuite with Matchers {
+
+  private val defaultTimeout = 200.millis
+
+  // longer than any test waits; a passing run never reaches it
```
mr-git:
```suggestion
  // longer than `ToTry` waits; a passing run never reaches it
```

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+    f.toTry(f.recovery) should matchPattern { case Failure(_: TimeoutException) => }
+
+    // what a retry around the flow does next
+    f.toTry(f.recovery) should matchPattern { case Failure(_: TimeoutException) => }
+
+    f.recorded shouldEqual Vector("acquired", "released", "acquired", "released")
+  }
+
+  test("an uncancelable guarded recovery completes and releases its guard") {
+    val f = fixture(timeout = 100.millis, recoveryDuration = 300.millis)
```
mr-git: How the `fixture(timeout = 100.millis, recoveryDuration = 300.millis)` differs from `fixture()`?

If it doesn't, then let's replace the AI-generated `class Fixture` and `def fixture()` pair (AI loves this pair, with over-loadable parameters, which are hard to reason about or follow) with something simpler, like `object Fixture` with `def apply()` inside it or even single instance of `Fixture`.

If it is required, then let's NOT have the defaults and let each spec define them, so that one can read every unit-test-case as a whole, without jumping down (to find the definition of `fixture`) and up (for definition of default value) to find which values are actually used!

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+
+    def recorded: Vector[String] = log.get()
+
+    /**
+     * Mimics `PartitionFlow`: take the guard and finish the acquire synchronously, then rebuild
+     * state across an asynchronous boundary. The release gives the guard back.
+     */
+    def recovery: IO[Unit] =
+      Resource
+        .make(guard.acquire *> record("acquired"))(_ => record("released") *> guard.release)
```
mr-git: in `skafka` we use `semaphore.permit.use(...).uncancelable`, not the pair of custom effects `.acquire` and `.release` - could you please add support for structure similar in `skafka`?

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
tobiajo: Lol

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
tobiajo: It is unlawful because `SyncIO` cannot actually be cancelled.

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
tobiajo: Simplified the instance, now it should be clear that `def rootCancelScope: CancelScope = CancelScope.Cancelable` is the only difference between `CancelableSyncIO` and `Sync[SyncIO]`.

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
tobiajo: The stepper stops at the outermost `uncancelable` and returns the entire block as the remainder, including the nesting. Thus leaves the rest to `IO.timeout` which comply with cats-effect's ordinary rules.

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
tobiajo: Changed to that shape.

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
tobiajo: The word soup is improved, keeping open until I verified OK.

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
tobiajo: I thought the fixtures were OK but I agree on the defaults.

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -35,31 +35,29 @@ object ToTry {
...
-            case Right(value) =>
-              Some(value)
-          }
-        }
-        a <- a.fold(error) { a => Success(a) }
-      } yield a
+    def apply[A](fa: IO[A]) = Try {
+      IO.asyncForIO.syncStep[SyncIO, A](fa, Int.MaxValue)(CancelableSyncIO).unsafeRunSync() match {
+        case Right(a) => a
+        case Left(rest) => rest.timeout(timeout).unsafeRunSync()
```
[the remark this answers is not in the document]
tobiajo: `scala.concurrent.TimeoutException` is a type alias for `java.util.concurrent.TimeoutException`.

tobiajo:
By the way, if we don't want either of #330 or #425, this is a workaround for kafka-flow: https://github.com/evolution-gaming/kafka-flow/pull/936

tobiajo:
The Codacy "issues" are false positives and misreading type, reacting on `Ref#get` thinking it is `Option#get`.

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+    val effect = IO.ref(0).flatMap { ref =>
+      (1 to 100_000).foldLeft(IO.unit)((acc, _) => acc.flatMap(_ => ref.update(_ + 1))) *> ref.get
+    }
+
+    toTry(effect) shouldEqual Success(100_000)
+  }
+
+  // `true`: per-record shapes (codecs, deserializers) that must stay on the calling thread.
+  // `false`: shapes the step must not enter because doing so drops protections. A fiber submission
+  // proves it stopped in front of them.
```
mr-git: very hard to read, feels like intro is missing. 

```suggestion
  // <SOMETHING HERE, like "The `callingThread` means"?>:
  //  - `true` - per-record shapes (codecs, deserializers) that must stay on the calling thread.
  //  - `false` - shapes which must not enter `WHAT???` because doing so drops protections. A fiber submission
  // proves it stopped in front of them.
```
what is `it`, which stopped in front of what `them`?

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+import com.evolutiongaming.catshelper.IOSuite._
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+/**
+ * `skafka` runs a Kafka rebalance callback through `ToTry[IO]`, and `kafka-flow` holds a semaphore
```
mr-git, line 14: The `PartitionFlow` in `kafka-flow` actually has `Resource.make(acquire) { _ => release }` shape, though I am not sure how it combines with `skafka`s `RebalanceListener1` usage.

Maybe more unit-test are required, I am not sure!

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  test("nested uncancelable: a timeout reaches a polled region and the permit comes back") {
+    val io = for {
+      guard <- Semaphore[IO](1)
+      cancelled <- Ref[IO].of(false)
+      // two uncancelable, both polled: the sleep stays cancelable
+      recovery = guard.permit.use { _ =>
+        IO.uncancelable { outer =>
+          outer(IO.uncancelable(inner => inner(IO.sleep(slowRecovery).onCancel(cancelled.set(true)))))
```
mr-git:
```suggestion
          outer(IO.uncancelable { inner => 
            inner(IO.sleep(slowRecovery).onCancel(cancelled.set(true))))
          }
```

mr-git (commented):
LGTM!

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
mr-git: I wonder why?

Honestly, I am a bit afraid that we might do here something unexpected :) 

On other hand - `ToTry` is broken at the edge of the timeout, this PR "unbreaks" it, somehow :)

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
```diff
@@ -35,31 +35,28 @@ object ToTry {
...
-              Some(value)
-          }
-        }
-        a <- a.fold(error) { a => Success(a) }
-      } yield a
+    def apply[A](fa: IO[A]): Try[A] = Try {
+      IO.asyncForIO.syncStep[SyncIO, A](fa, Int.MaxValue)(CancelableSyncIO).unsafeRunSync() match {
+        case Right(a) => a
+        case Left(rest) => rest.timeout(timeout).unsafeRunSync()
+      }
```
nviliunov-evolution-throwaway, line 59:
```suggestion
      fa.timeout(timeout).unsafeRunSync()
```

This will make all of your tests pass. The Fiber execution already handles all possible sync sections on the current thread. Non need to reinvent the wheel.

nviliunov-evolution-throwaway: One could say that this solution is less performant, but 1) do we actually have evidence that motivated the initial solution? 2) as far as I am aware, in most cases we will call `unsafeRunSync()` anyway and pay that overhead, it doesn't matter whether we do it at the start or in the middle.

tobiajo: I did this little comparison: https://github.com/tobiajo/cats-helper/pull/3

There is less overhead on the synchronous path. It may make a difference if `ToTry` is abused for like per-record deserialization, which is exactly what skafka does. Also kafka-journal goes through `ToTry` per event.

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
tobiajo: Quite equivalent testing with semaphore resource here I would say.

core/src/test/scala/com/evolutiongaming/catshelper/ToTrySyncStepSpec.scala
tobiajo: Addressed

core/src/test/scala/com/evolutiongaming/catshelper/ToTryTimeoutSpec.scala
tobiajo: Applied

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
tobiajo: I'm happy with either, someone just need to decide which it is.

nviliunov-evolution-throwaway: Ok, the comparison is very persuasive. I am convinced that this fix works, and that it using direct `unsafeRunSync` would have a drastic performance penalty. However, the fix in this PR as far as I'm concerned depends on implementation details of `SyncStep.interpret`, which may change in the future. This is not sustainable in the long term, as consumers of `cats-helper` are free to bump `cats-effect` independently.

We can merge it, but only as a hotfix, with intention to introduce more robust fixes, of which I see a few potential ones:
1) Kafka-Flow `ToBytes`/`FromBytes` [de]serializers should not have used `IO` in the first place, it's fine to have `Try` as `F` there. Unless it's the same interface that does schema registry requests.
2) `syncStep` seems to forget finalizers when walking through specifically `OnCancel`. I'd imagine it would somehow collect finalizers into the resulting `SyncIO[IO[T] | T]`. We should raise it with cats-effect maintainers, and if it's impossible to do so, `syncStep` should be document as a lossy interpreter and potentially deprecated.
3) A possibly more significant change, but `unsafeRunSync` or a new similar method could avoid spawning a fully-fledged fiber onto the provided executor and instead run it straight on the current thread until a first cede. This is in line with how cats-effect dynamically treats execution threads, and since current implementation blocks the current thread, we get thread-saving for free. It's important that `unsafeRunSync` and `toTry.apply` share the IO interpretation, otherwise we'll get such divergences.

tobiajo: I will open an upstream issue today.

tobiajo: Here it is:
- https://github.com/typelevel/cats-effect/issues/4687

core/src/main/scala/com/evolutiongaming/catshelper/ToTry.scala
tobiajo: Changed the wording, "unlawful" was too strong. `SyncIO`'s `canceled` is a no-op, same as `unit`.

